PR #47, opened 2012-11-20
Ability to manage email subjects from configuration.

doobeh:
Let me know if I went down the wrong track with this.


TimotheeJeannin:
Is there any way to translate the email subjects ?

Issue #155, opened 2013-08-19
make _default_field_labels configurable


edelans:
It would be a nice feature for non english-speaking users.

Issue #18, opened 2012-06-18
Consider Flask-Babel integration


mattupstate:
i10n and i18n could be necessary for things such as Flash messages.

yoursdearboy:
In addition to Chris solution - to localize login message `SECURITY_MSG_LOGIN` it's also necessary to provide localize callback for login manager.

```python
security = Security(app, user_datastore)
security.login_manager.localize_callback = lazy_gettext
```

jirikuncar:
@yoursdearboy the Flask-BabelEx has been integrated in #617 (originally from https://github.com/inveniosoftware/flask-security-fork/pull/33). Please check if it works for you and in case you have a problem open new issue. Thank you!

PR #203, opened 2014-01-03
Allows field labels to be set in config

maebert:
This makes it relatively easy to e.g. localise them, by simply in your app config saying

```python
from flask.ext.babel import lazy_gettext as _

SECURITY_LABEL_EMAIL = _("Email Address")
SECURITY_LABEL_PASSWORD = _("Password")
```

I guess the problem with #155 was that the forms classes were constructed outside an application context. I solved that by creating a proxy object that holds back the actual string realisation until it's needed.

This makes #18 also relatively trivial now.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -126,6 +126,22 @@ _default_messages = {
     'REFRESH': ('Please reauthenticate to access this page.', 'info'),
 }
 
+_default_field_labels = {
+    'EMAIL': 'Email Address',
+    'PASSWORD': 'Password',
+    'REMEMBER_ME': 'Remember Me',
+    'LOGIN': 'Login',
+    'RETYPE_PASSWORD': 'Retype Password',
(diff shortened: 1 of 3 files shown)


mattupstate:
@maebert Not sure if you're still interested in this PR but I'll be happy to accept if you can merge the upstream updates.

felix1m:
i manually applied this patch to the current develop branch since i needed it myself. Can confirm it is working properly. Note: When using non-asci characters no not forget to use u'öäü' in python 2.x.

https://github.com/felix1m/flask-security/commit/da8c391b05fa3eddfd67bf8e0bf69e7c05c9698f

Should i create a seperate PR for this (since it is not really my work :) )?

maebert:
@felix1m Feel free to post a PR, it's probably less work than rebasing it!

felix1m:
see pr #390

PR #246, opened 2014-05-05
Added option for HTML and PLAINTEXT selection for emails

JinBlack:
I Added options SECURITY_EMAIL_PLAINTEXT and SECURITY_EMAIL_HTML to set the type of email to be sent. You can disable one of them or have both as before this patch.


mattupstate:
@JinBlack if you can merge upstream updates I'll accept this.

PR #255, opened 2014-05-28
Allow overriding of an unauthorized callback.

nfvs:
Right now, when using `@http_auth_required` or `@auth_token_required`, if the user is unauthorized only a message is returned (`_default_unauthorized_html`).

With this change, it is possible to define an unauthorized callback which will be called instead.

Example using your overholt project (`overholt/api/__init__.py`):

```python
from overholt.core import security

def unauthorized():
    return jsonify(dict(error='Unauthorized')), 403

def create_app(settings_override=None, register_security_blueprint=False):
    app = factory.create_app(__name__, __path__, settings_override,
                             register_security_blueprint=register_security_blueprint)
    security.unauthorized_handler(unauthorized)

```

I have tested it only with `@http_auth_required`, but `@auth_required` and `@auth_token_required` should work pretty much the same.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -273,7 +273,8 @@ def _get_state(app, datastore, **kwargs):
         reset_serializer=_get_serializer(app, 'reset'),
         confirm_serializer=_get_serializer(app, 'confirm'),
         _context_processors={},
-        _send_mail_task=None
+        _send_mail_task=None,
+        _unauthorized_callback=None
     ))
 
     for key, value in _default_forms.items():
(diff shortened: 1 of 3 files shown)


mattupstate:
@nfvs I'm happy to accept this if you can merge upstream changes.

nfvs:
It lives! :-)

Had some troubles with the rebase, which is why there are so many messages after your last one, but it should be done now. I can also squash them into a single commit if you'd like.

mattupstate:
@nfvs Yeah! Sorry for being so slow. Life gets in the way of open source sometimes. Squashing would be great too.

nfvs:
Absolutely, no worries. Squashed.

nfvs:
Had to rebase again since you just merged the WWW-Authenticate PR, which had conflicts with this one (thanks btw).

PR #262, opened 2014-06-06
Custom request loader

ocean1:
Flask-login allows authenticating every request through a request loader (i.e. rest API auth)
and added the request loader, the api key generation is for now left to the user (i implemented it as hash of the concatenation of user email and concatenation of random/secret data) and it could be added to the UserMixin class.

there's also this p.r. for flask-login https://github.com/maxcountryman/flask-login/pull/163 to avoid having unneeded cookies, at the moment is a global config, but it could be hacked to ensure that if it's a request for the request loader session can be cleaned, but would require more effort.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -226,7 +226,7 @@ def _request_loader(request):
         if user is not None:
             return user
     except Exception as e:
-        print e
+        pass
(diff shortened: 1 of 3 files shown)


mattupstate:
I'm afraid this is not a feature I want to support.

PR #288, opened 2014-07-29
Use EmailField from wtforms.fields.html5 for email fields in forms

scollinson:


mattupstate:
I'm afraid I'm not concerned with HTML5 support.

Issue #194, opened 2013-12-12
Document send_mail_task


nickretallack:
I was hoping to make my app faster at sending confirmation emails when I noticed [this](https://github.com/mattupstate/flask-security/blob/develop/flask_security/utils.py#L253) in the code.  Could you add a bit of explanation as to how to use this?  I assume it has something to do with [celery](http://flask.pocoo.org/docs/patterns/celery/).

erwagasore:
```
from app.factory import create_celery_app, mail, security

celery = create_celery_app()

@celery.task()
def send_security_mail(msg):
    mail.send(msg)

@security.send_mail_task
def delay_security_mail(msg):
    send_security_mail.delay(msg)
```

@mattupstate, based on the code above I am getting a ```RuntimeError: maximum recursion depth exceeded``` when I run celery ```celery worker -A app.tasks -l INFO```

joaobarbosa:
@erwagasore did you managed to fix it? I'm having the same issue :cry:

eriktaubeneck:
I think this is likely the same issue [described in issue 340](https://github.com/mattupstate/flask-security/issues/340).

joaobarbosa:
Thanks @eriktaubeneck!

jxltom:
@erwagasore @joaobarbosa @eriktaubeneck 

If you are going to use factory method, recursion error will raise if it is not used properly. Here is the example based on author's example for your reference. The trick is using the ```state``` object returned by ```init_app``` for setting cutomized mail sending function. The author has explained it [here](https://github.com/mattupstate/flask-security/issues/141#issuecomment-21728482)  and [here](https://github.com/mattupstate/flask-security/issues/141#issuecomment-21744531) in #141.

```python
from celery import Celery
from flask import Flask
from flask_security import Security, .. datastore stuff ..
from flask_mail import Mail

celery = Celery()
mail = Mail()
security = Security()

def create_app():
    app = Flask(__name__)
    mail.init_app(app)
    security_ctx = security.init_app(app,..datastore stuff..)

    @celery.task
    def send_flask_mail(msg):
        mail.send(msg)

    # Flexible way for define custom mail sending function.
    @security_ctx.send_mail_task
    def delay_flask_security_mail(msg):
        send_flask_mail.delay(msg)

    # You can also define like this as shortcurt
    security_ctx.send_mail_task(send_flask_mail.delay)

    return app
```

jirikuncar:
[...]

PR #317, opened 2014-09-25
strange recursion error, when using Security() and later init_app()

joergwork:
One can't do something like:

`my_security = Security()`
and later on
`my_security.init_app(app, ...)`

'cause that ends in a strange recursion. 
(Haven't tested that, but it should work. This was an issue on #pocoo from ianseyer)

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -392,7 +392,7 @@ class Security(object):
         self.datastore = datastore
 
         if app is not None and datastore is not None:
-            self._state = self.init_app(app, datastore, **kwargs)
+            self.init_app(app, datastore, **kwargs)
 
     def init_app(self, app, datastore=None, register_blueprint=True,
                  login_form=None, confirm_register_form=None,
@@ -432,8 +432,8 @@ class Security(object):
 
         state.render_template = self.render_template
         app.extensions['security'] = state
-
-        return state
+        
+        self._state = state
 
     def render_template(self, *args, **kwargs):
         return render_template(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


Trii:
I ran into the same issue today because of self._state not being set. I can confirm that this fix works.

```python
# Import the Flask Framework
from flask import Flask
from flask_security import Security
from flask_security_ndb import NDBUserDatastore, User, Role

# Flask app
app = Flask(__name__)

# Setup Flask-Security using the NDB adapter WIP
user_datastore = NDBUserDatastore(User, Role)
security_app = Security(datastore=user_datastore)

security_app.init_app(app)

# Override Flask-Mail for Google App Engine by 
# using the hook in :class:`flask_security.core._SecurityState`
security_app.send_mail_task(send_email)
```

russomi:
+1

russomi:
Any thoughts on this one?  :neckbeard:


---

Reviewed 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mattupstate/flask-security/317)

davvid:
I have a simpler fix in my fork that addresses this in a better way.  We shouldn't set _state on the instance when using the deferred init_app() factory pattern.  Instead, provide a default local proxy so that the existing getattr code works as-is.  I'll open a new PR with that sug.

jxltom:
I have posted a example on how to initialize properly using factory method [here](https://github.com/mattupstate/flask-security/issues/194#issuecomment-278554575] in #194. The key is using ```state``` object returned by ```init_app``` instead of ```Security``` instance.

PR #322, opened 2014-10-01
Add configuration for token expiration

waltaskew:
Add the ability to set token expiration by configuring the SECURITY_TOKEN_MAX_AGE value.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,3 +36,4 @@ Rotem Yaari
 Srijan Choudhary
 Tristan Escalada
 Vadim Kotov
+Walt Askew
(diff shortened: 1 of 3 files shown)


doender:
+1

mattupstate:
Thanks for this.

pdonorio:
I was trying to use the "SECURITY_TOKEN_MAX_AGE" found here with the pull of latest code.

Using "1 min":
The token is new at login after one minute, but the old tokens still work.
Am i doing something wrong?

jonafato:
@pdonorio per the docs included, this should be number of seconds (an integer, not a human-readable string).

pdonorio:
> this should be number of seconds (an integer, not a human-readable string)

Tested, and you are right. Thanks @jonafato !

galvarado:
Hi, this feature is not available on  the current Version (1.7.4),  right? When will be another release? The newest was on October 13th 2014. Many things have been improved since that day, as the flask-login backward compatibility and the token expiration, but as far as I see all the commits and PR's are  on develop branch.Thanks for your time @mattupstate . (+1 For this excellent plugin)

PR #327, opened 2014-10-09
Committing changes in SQLAlchemyDatastore

Xumeiquer:
When performing any action which modifies the data base like create_role or add_role_to_user, for example. The modifications wasn't committed into the database, so adding `self.commit()` in `SQLAlchemyDatastore` under the `put` and `delete` functions solves the issue.

### DIFF
--- a/flask_security/datastore.py
+++ b/flask_security/datastore.py
@@ -32,10 +32,12 @@ class SQLAlchemyDatastore(Datastore):
 
     def put(self, model):
         self.db.session.add(model)
+        self.commit()
         return model
 
     def delete(self, model):
         self.db.session.delete(model)
+        self.commit()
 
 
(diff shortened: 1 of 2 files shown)


jonafato:
I can see where this may be useful, but I'm not a fan of implicit commits like this. SQLAlchemy encourages explicit session commits, and many people expect and depend on their code to work that way. All of Flask-Security's views that modify data already commit using `after_this_request`, and elsewhere, a developer probably doesn't want Flask-Security committing their session under the hood before all of the modifications to that session have been made.

Is there a compelling reason this should happen within the `put` and `delete` methods or a use case where this makes code significantly cleaner?

jonafato:
I'm closing this due to lack of response. If there is still interest in elaborating on the rationale behind this PR, feel free to reopen it.

PR #335, opened 2014-11-05
Fix script. Add commands and a Sub-Manager

fmerges:
I noticed script.py, not sure if this is deprecated or still in use...

Having the sub-manager will allow to do something like this in your main manage.py:

from flask.ext.security.script import manager as security_manager
manager.add_command('security', security_manager)

### DIFF
--- a/flask_security/script.py
+++ b/flask_security/script.py
@@ -18,13 +18,16 @@ except ImportError:
 import re
 
 from flask import current_app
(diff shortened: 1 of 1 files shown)


mattupstate:
This could potentially break things for existing users. Additionally, I'd like to revisit the commands when Flask ships with its new CLI tools.

PR #351, opened 2014-12-21
Add documentation about SECURITY_MSG configuration variables.

denizdogan:
I had trouble figuring out how to override the default messages, so I added some information to the configuration documentation about SECURITY_MSG configuration variables.


mattupstate:
Thanks for this!

PR #354, opened 2015-01-07
Fix #353, if user use an old access token, then the user will be None

### LINKED ISSUE #353: Check whether the user is None in _check_user() in decorator.py
If you use a valid access token of an user, but this user is not in the database right now. Then the user will be None, and there will be an exception if you access it's password later.

<code>
 user = _security.login_manager.token_callback(token)
<code>

ychiaoli18:
if user use an old access token, then the user will be None, so we have to return False if the user is None

### DIFF
--- a/flask_security/decorators.py
+++ b/flask_security/decorators.py
@@ -55,6 +55,8 @@ def _check_token():
         token = request.json.get(args_key, token)
 
     user = _security.login_manager.token_callback(token)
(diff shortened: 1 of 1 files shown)


mattupstate:
Can provide some sort of test illustrating the scenario you speak of?

jonafato:
Closing due to lack of response.

PR #361, opened 2015-01-30
Include WWW-Authenticate headers in @auth_required.

nfvs:
When using @http_auth_required, the WWW-Authenticate header is included, but when using @auth_required('basic'), it is not.

This change includes that header in every @auth_required call that contains the 'basic' method.


mattupstate:
Thanks!

PR #366, opened 2015-02-18
Modified check_token function to handle list-type JSON post

covertgeek:
This modifies the check_token function to resolve an issue encountered when posting multiple objects in a JSON request.  The library expects a dictionary and causes the server to throw a 500-error when it encounters a list instead.

The patch assumes that if request.json is a list, that the authentication token will be in the header.


mattupstate:
If you merge upstream changes I can accept this.

covertgeek:
@mattupstate I apologize for the long delay in responding to this. I've merged the upstream changes from master and pushed them back up. Is there anything I need to do at this point or is it automatic?

mattupstate:
Thanks!

PR #376, opened 2015-03-24
Encrypt raw password, not HMAC of password.

### LINKED ISSUE #320: having hmac required makes supporting legacy passwords w/o hmac impossible
i've migrated user accounts from an old django app. even though I can set schemes correctly (with the undocumented SECURITY_PASSWORD_SCHEMES) like so:

```
SECURITY_PASSWORD_HASH = 'sha256_crypt'
SECURITY_PASSWORD_SCHEMES = ['sha256_crypt', 'django_pbkdf2_sha256'] 
```

old passwords won't verify because the function utils.verify_and_update_password first gets the hmac version of the password before verifying through passlib. this seems odd, a key features of passlib is the ability to support multiple encryption schemes and legacy passwords. am I missing something?

longhotsummer:
This passes passwords directly to passlib's CryptContext
methods and relies on it to do the hard work.

For backwards compatibility we fall back to trying the old HMAC approach
when verifying passwords, but only if a salt has been set. Legacy passwords
are updated upon check.

Previously, we passed the HMAC'd password to passlib, which meant that
we weren't compatible with other users of passlib (such as Django) or
the Modular Crypt Format at
https://pythonhosted.org/passlib/modular_crypt_format.html

Fixes #320

### DIFF
--- a/flask_security/__init__.py
+++ b/flask_security/__init__.py
@@ -10,8 +10,6 @@
     :license: MIT, see LICENSE for more details.
 """
 
-__version__ = '1.7.4'
-
 from .core import Security, RoleMixin, UserMixin, AnonymousUser, current_user
 from .datastore import SQLAlchemyUserDatastore, MongoEngineUserDatastore, PeeweeUserDatastore
 from .decorators import auth_token_required, http_auth_required, \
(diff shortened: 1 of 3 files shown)


longhotsummer:
I'd like to get feedback on this before going through and updating the docs.

PR #373, opened 2015-03-06
Add support for custom AnonymousUser class.

lnielsen:

PR #377, opened 2015-03-25
fix pep8 import error

longhotsummer:
Fixes a pep8 error which is breaking a number of pull request builds.


longhotsummer:
Hi @mattupstate -- any feedback on this? Thanks!

jonafato:
Thanks for the fix. This change has already been pulled in as a result of #373's merge.

PR #378, opened 2015-03-25
Allow users to override verify_password, verify_and_update_password and ...

longhotsummer:
...encrypt_password.

This is a different approach to fixing #320 in that it allows users to
sub in their own implementations of these functions. As it currently
stands, there's no way to do so.


rturk:
any progress on this?

mcouthon:
This would be really useful.

jirikuncar:
@longhotsummer We have added few more configuration options for password verification without double hashing. Please re-open it with test cases in case you still need it. Thank you!

Issue #375, opened 2015-03-24
Reset password error when user only has Flask-Social login


donnywals:
I'm not sure how this should be resolved because the password seems to be a hard dependency for the reset functionality but when a user logs in to an app with a social account (Using Flask-Social), an account is created for them. This account has an email address but no password. So when you go to the /reset page and enter the email address it throws an error: 

TypeError: must be string or buffer, not None

Debugging this shows that the md5 function is throwing this error, most likely because the user password doesn't exist? Does anybody have suggestions?

nickretallack:
I'm also having this issue because I'm using an invitation system.  If a user attempts to reset their password before setting their password in the invitation workflow, it throws an error.

nickretallack:
I looked at the code and it turns out the password md5 there is currently totally inconsequential.  You can take it out and it runs just fine.  So I made it ok with null passwords.  But then I realized what you really want is for it to invalidate old password reset requests when your password changes, so I actually made use of the password md5 if it's there.  Enjoy.

jonafato:
This is fixed by #389

PR #382, opened 2015-04-08
Reset Password bug fixes / enhancements

nickretallack:
Fixes https://github.com/mattupstate/flask-security/issues/375 and makes the reset token actually expire once it is used.

### DIFF
--- a/flask_security/recoverable.py
+++ b/flask_security/recoverable.py
@@ -63,11 +63,17 @@ def reset_password_token_status(token):
     """Returns the expired status, invalid status, and user of a password reset
     token. For example::
 
-        expired, invalid, user = reset_password_token_status('...')
+        expired, invalid, user, data = reset_password_token_status('...')
 
     :param token: The password reset token
     """
-    return get_token_status(token, 'reset', 'RESET_PASSWORD')
+    expired, invalid, user, data = get_token_status(token, 'reset', 'RESET_PASSWORD', return_data=True)
+    if not invalid:
+        password_hash = md5(user.password) if user.password else None
+        if password_hash != data[1]:
+            invalid = True
+
+    return expired, invalid, user
 
 
 def update_password(user, password):
--- a/flask_security/recoverable.py
+++ b/flask_security/recoverable.py
@@ -54,7 +54,8 @@ def generate_reset_password_token(user):
 
     :param user: The user to work with
     """
-    data = [str(user.id), md5(user.password)]
+    password_hash = md5(user.password) if user.password else None
+    data = [str(user.id), password_hash]
     return _security.reset_serializer.dumps(data)
 
 
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -333,7 +333,7 @@ def send_mail(subject, recipient, template, **context):
     mail.send(msg)
 
 
-def get_token_status(token, serializer, max_age=None):
+def get_token_status(token, serializer, max_age=None, return_data=False):
     """Get the status of a token.
 
     :param token: The token to check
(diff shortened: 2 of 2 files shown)


nickretallack-business:
Can this be accepted added soon?

nickretallack:
Can this be accepted soon?

mattupstate:
@nickretallack I'm not sure I understand this so-called enhancement. Nor do I see any tests that illustrate the use case.

nickretallack-business:
Details are in the linked issue.  I'll look into how to test it.

The primary issue is, if you attempt to log in as a user who has no password set, you'll raise an exception.

nickretallack:
Details are in the linked issue.  I'll look into how to test it.

The primary issue is, if you attempt to log in as a user who has no password set, you'll raise an exception.

jonafato:
It seems to me that this pull request addresses two unrelated issues.

1. Currently, the MD5 has of a user's current password is used in the generation of a password reset token. When a user has no password (to pull some reasons from the linked issue: a user only currently authenticated via some social API or an account created on behalf of some user as part of an invite-only system), the password can be `None` causing the MD5 step to raise an exception. I agree that this is an issue that needs to be solved.

2. Additionally, the PR is adding explicit logic to consider a password reset token to be invalid if the MD5 of the user's current password hash is not equal to the value pulled from the token. I should note that this is a breaking change if anyone is relying tokens working more than once within their max_age lifetime. This is not a good thing to depend on, but it should make its way into the changelog if accepted.

Both of these changes seem like welcome additions, but they should probably be submitted as separate changes and should definitely be accompanied by tests.

mattupstate:
Users without passwords is strange to me. If a user is created without a password then it is most likely being done outside the context of Flask-Security code. If this is necessary, why would one not create a randomly generated password for the user when the user is created to avoid all this?

mattupstate:
@nickretallack you're going to have to forgive me for being so short sighted. In all honesty, I always thought this was the way I had implemented it. Clearly I'm a moron. What you've contributed here is what I always intended to have, so let me thank you for the additions and again apologize for being such a pain about your PR.

@jonafato do you know of anyone that expects password reset tokens to work multiple times?

jonafato:
I meant to post the following earlier today but closed the tab.

@mattupstate A concrete example: for PyGotham, we occasionally create accounts on behalf of invited speakers. In the interest of not sending a randomly generated password in cleartext, we send them a password reset link. Our current workaround for the `MD5(None)` issue is to set the password hash to some value that a legitimate hash will never collide with and generate the reset link based on that. This is roughly equivalent to what you suggest, but it would be convenient if it was not a required step.

The other change ensures that a password reset link cannot be used after the hash with which it was generated has changed. This is not currently the case, and I think it's a necessary step. If someone does currently expect a password reset token to work more than once, they're explicitly weakening the security of their application and should be discouraged / prevented from doing so. That said, it's an important change in behavior, and my main point was that this should make it into the changelog.

mattupstate:
Understood! Now we just need some tests.

flask_security/recoverable.py
```diff
@@ -62,11 +63,17 @@ def reset_password_token_status(token):
...
-        expired, invalid, user = reset_password_token_status('...')
+        expired, invalid, user, data = reset_password_token_status('...')
 
     :param token: The password reset token
     """
-    return get_token_status(token, 'reset', 'RESET_PASSWORD')
+    expired, invalid, user, data = get_token_status(token, 'reset', 'RESET_PASSWORD', return_data=True)
+    if not invalid:
+        password_hash = md5(user.password) if user.password else None
+        if password_hash != data[1]:
```
mattupstate: should be using `werkzeug.security.safe_str_cmp` here.

nickretallack:
New pull request: https://github.com/mattupstate/flask-security/pull/389#issuecomment-101155709

PR #384, opened 2015-04-15
Fix inconsistent behavior between Security(..) and init_app(..)

### LINKED ISSUE #383: Inconsistent behavior between Security(...) and init_app(...)
Security(app, datastore)

stores the app and datastore in the object (`self`)

    security = Security()
    [...]
    security.init_app(app, datastore)

does not. This causes the `security.datastore` to be `None` and thus inaccessible.

Diaoul:
Fixes #383

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -389,7 +389,7 @@ class Security(object):
     """
     def __init__(self, app=None, datastore=None, **kwargs):
         self.app = app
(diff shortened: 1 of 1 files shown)


KeNaCo:
@Diaoul hi, may I ask, what is the status of this PR? Are you try to fix failing tests or is this dead?

PR #397, opened 2015-06-23
Change flask.ext.foo imports to flask_foo

jonafato:
flask.ext has been [deprecated](http://flask.pocoo.org/docs/dev/extensiondev/#ext-import-transition), and users are now encouraged to
import using the explicit package name. This change updates docs to
reflect that.


jonafato:
Fixed by #528.

PR #398, opened 2015-06-23
Restrict bcrypt to <2.0.0

jonafato:
As of 2.0.0, passlib no longer correctly identifies bcrypt as bcrypt
(instead, it mistakenly applies pybcrypt logic to bcrypt). This results
in all Python 3 logic involving bcrypt failing. As a hotfix, we should
require users to be on a version of bcrypt that passlib can handle a fix
can be pushed into passlib.


jonafato:
I've filed a [related issue](https://code.google.com/p/passlib/issues/detail?id=56) on Google code.

mattupstate:
LGTM

PR #399, opened 2015-06-26
add SECURITY_POST_LOGIN_ALLOW_REDIRECT_DOMAIN option

wong2:
Consider this: after login from `www.domain.com`, the user should be redirected to `en.domain.com`, which is not allowed by Flask-Security currently.

so a new option `SECURITY_POST_LOGIN_ALLOW_REDIRECT_DOMAIN` is introduced, in the above situation, one could set `SECURITY_POST_LOGIN_ALLOW_REDIRECT_DOMAIN = '*.domain.com'` to allow such redirection.


wong2:
ping @mattupstate

docs/configuration.rst
```diff
@@ -314,6 +314,11 @@ Miscellaneous
 ``SECURITY_DATETIME_FACTORY``                 Specifies the default datetime
                                               factory. Defaults to
                                               ``datetime.datetime.utcnow``.
+``POST_LOGIN_ALLOW_REDIRECT_DOMAIN``          Specifies domain allowed to be
```
jirikuncar, line 317: Please use `SECURITY_` prefix.

docs/configuration.rst
```diff
@@ -314,6 +314,11 @@ Miscellaneous
...
+                                              redirected to after login. You can
+                                              use wildcard patterns like 
+                                              ``*.doumail.com``. Defaults to the
```
jirikuncar, line 320: ``*.doumail.com`` -> ``*.example.com`` to be clear that it's an example.

wong2:
sorry it has been too long, can't remember the details here

PR #402, opened 2015-07-14
Test and fix script commands

wjt:
I wired up the commands from `flask.ext.security.script`, and discovered:
- `ActivateUserCommand` and `DeactivateUserCommand` don't work at all, as far as I can tell
- None of them fail gracefully if required parameters aren't supplied
- Passing `--active` as a flag to `CreateUserCommand` is wrong, you have to pass one of a list of undocumented parameters to mean "True"
- There are no tests

This PR fixes all except the first issue. If I get a chance I'll fix `_ToggleActiveCommand` too.

### DIFF
--- a/flask_security/datastore.py
+++ b/flask_security/datastore.py
@@ -133,6 +133,7 @@ class UserDatastore(object):
         user = self._prepare_active_modify_args(user)
 
         user.active = not user.active
+        self.put(user)
         return True
 
     def deactivate_user(self, user):
@@ -144,6 +145,7 @@ class UserDatastore(object):
 
         if user.active:
             user.active = False
+            self.put(user)
             return True
         return False
 
@@ -156,6 +158,7 @@ class UserDatastore(object):
 
         if not user.active:
             user.active = True
+            self.put(user)
             return True
         return False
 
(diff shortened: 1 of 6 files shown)


wjt:
Fixed `ActivateUserCommand` and `DeactivateUserCommand` as well.

I don't know if there was a good reason why the peewee datastore in the test suite required username to be non-null to, whereas the mongo and sqlalchemy ones don't. The script commands identify users by email address and do not support passing a username. So making that field nullable fixes most of the tests with peewee. I don't know what's going on with the one that still fails with peewee.

wjt:
Fixed (and wrote more tests) for the failures when modifying `User.active` with Mongo and Peewee. I also added one more fix + test for `script.py`, bringing it up to 100% coverage.

wjt:
I guess the changes to the Mongo and Peewee backends are related to #327 – it seems like implicit commits are already the norm for those backends so my change here seems consistent (though I agree with the dissent on that PR).

jirikuncar:
@wjt please check new Click interface implemented in 93ae74b679fec7b58a110611a5f7cfe1ae23a423.

PR #406, opened 2015-07-28
sets form validation messages upon instantiation

dlakata:
Currently, the SECURITY_MSG error messages only affect the form's validation feedback after it's been validated. However, for JavaScript libraries that access the form's validators before the form is actually submitted, the error messages that are shown are the configuration names of the messages, not the actual messages themselves. This commit sets the appropriate SECURITY_MSG messages upon instantiation of the form.


jirikuncar:
Please re-open the PR if it is still needed with a more details on how to use it. Thanks!

PR #408, opened 2015-07-29
removes duplicate label in _default_field_labels of forms

dlakata:


jirikuncar:
@dlakata thanks! It have been fixed meantime.

PR #409, opened 2015-07-31
Fixing initialization kwargs.

sdorazio:
Currently, the args passed to `init_app` & `_get_state` are overwritten by the app / default config values. The `kwargs` should take a higher precedence.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -260,7 +260,8 @@ def _get_serializer(app, name):
 
 def _get_state(app, datastore, anonymous_user=None, **kwargs):
     for key, value in get_config(app).items():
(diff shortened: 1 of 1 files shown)


flask_security/core.py
```diff
@@ -260,7 +260,8 @@ def _get_serializer(app, name):
 
 def _get_state(app, datastore, anonymous_user=None, **kwargs):
     for key, value in get_config(app).items():
-        kwargs[key.lower()] = value
+        if kwargs.get(key.lower()) is None:
```
jirikuncar, line 263: Can we simply use `kwargs.setdefault(key.lower(), value)`?

PR #411, opened 2015-08-06
Added IGNORE_STATIC_ENDPOINTS option

a-pertsev:
Added option IGNORE_STATIC_ENDPOINTS that will be proxied to Principal extension.
Sometimes it is usefull (for example, if we want to have some heavy hook on identity load or request start).


jirikuncar:
I am closing it in favor of https://github.com/mattupstate/flask-principal/pull/49

PR #417, opened 2015-09-11
Fixes #416 Changing is_authenticated from function to property & updating Flask-…

ialex:
…Login>=0.3.0

PR #418, opened 2015-09-14
change is_authenticated and is_active from method to property

guoqiao:
Flask-Login 3.0 has change all is_xxx method to property, which is not compatible. Need to update related code.

### DIFF
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -195,7 +195,7 @@ class PasswordlessLoginForm(Form, UserEmailFormMixin):
     def validate(self):
         if not super(PasswordlessLoginForm, self).validate():
             return False
-        if not self.user.is_active():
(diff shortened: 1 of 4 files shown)


mattupstate:
This is a duplicate of #417

flask_security/forms.py
```diff
@@ -195,7 +195,7 @@ def __init__(self, *args, **kwargs):
     def validate(self):
         if not super(PasswordlessLoginForm, self).validate():
             return False
-        if not self.user.is_active():
+        if not self.user.is_active:
```
ialex: in #417 this is not changed but test are passing is it necesary to also change this with Flask-Login 0.3.0? maybe these lines are not covered by any test. i can change the PR if required

mattupstate:
Fixed by #438

PR #423, opened 2015-09-23
master branch does not support Flask-Login >=0.3

fuhrysteve:
If there is a minor release prior to resolving the issue upstream in #417, I believe this is the proper solution that does not break backwards compatibility.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Flask>=0.9
-Flask-Login>=0.1.3
+Flask-Login>=0.1.3,<0.3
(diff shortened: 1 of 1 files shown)


stinaq:
I have run in to this error. By just doing pip install Flask-Security at the moment it will not run. I got Flask-Login v 0.3.2 by default, which will not work, so I had to set Flask-Login to 0.2.11
I can see that the fix is in the develop branch, will it come to release soon?

mattupstate:
Fixed by #438

PR #425, opened 2015-09-24
remove dependency on flask-sqlalchemy

nitishr:
It seems better to remove the dependency on flask-sqlalchemy if we can achieve exactly the same functionality without it, using plain old sqlalchemy.

Additionally, the tests now use sqlite in memory instead of file, making the test suite many times faster.

This change is currently backward incompatible, since SQLAlchemyUserDatastore now takes a  session from sqlalchemy instead of SQLAlchemy from flask-sqlalchemy. So, SQLAlchemyUserDatastore(db, User, Role) now becomes SQLAlchemyUserDatastore(db.session, User, Role). Can be made easily backward compatible, though, by adding an isinstance check (preferable), or creating a new SQLAlchemySessionUserDatastore.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -15,7 +15,7 @@ SQLAlchemy Install requirements
 ::
 
      $ mkvirtualenv <your-app-name>
-     $ pip install flask-security flask-sqlalchemy
+     $ pip install flask-security sqlalchemy
 
 
 SQLAlchemy Application
@@ -60,7 +60,7 @@ possible using SQLAlchemy:
                                 backref=db.backref('users', lazy='dynamic'))
 
     # Setup Flask-Security
-    user_datastore = SQLAlchemyUserDatastore(db, User, Role)
+    user_datastore = SQLAlchemyUserDatastore(db.session, User, Role)
     security = Security(app, user_datastore)
 
     # Create a user to test with
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -150,44 +150,45 @@ def mongoengine_datastore(request, app):
 
 @pytest.fixture()
 def sqlalchemy_datastore(request, app, tmpdir):
-    from flask_sqlalchemy import SQLAlchemy
+    from sqlalchemy import Table, Column, Integer, ForeignKey, String, DateTime, Boolean, MetaData, create_engine
+    from sqlalchemy.ext.declarative import declarative_base
+    from sqlalchemy.orm import relationship, backref, sessionmaker
 
-    f, path = tempfile.mkstemp(prefix='flask-security-test-db', suffix='.db', dir=str(tmpdir))
+    metadata = MetaData()
+    Model = declarative_base(metadata=metadata)
 
-    app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///' + path
-    db = SQLAlchemy(app)
+    roles_users = Table(
+        'roles_users', metadata,
+        Column('user_id', Integer(), ForeignKey('user.id')),
+        Column('role_id', Integer(), ForeignKey('role.id')))
 
-    roles_users = db.Table(
-        'roles_users',
-        db.Column('user_id', db.Integer(), db.ForeignKey('user.id')),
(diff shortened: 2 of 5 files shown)


docs/quickstart.rst
```diff
@@ -60,7 +60,7 @@ possible using SQLAlchemy:
                                 backref=db.backref('users', lazy='dynamic'))
 
     # Setup Flask-Security
-    user_datastore = SQLAlchemyUserDatastore(db, User, Role)
+    user_datastore = SQLAlchemyUserDatastore(db.session, User, Role)
```
fuhrysteve: I would probably add a comment here that says something like `if using flask-sqlalchemy, use db instead of db.session`

fuhrysteve:
I like this idea - I don't really see any upside to depending on `Flask-SQLAlchemy`, although maybe it brings something to the table here that I'm not aware of.

At any rate, better to leave things decoupled and make fewer assumptions when possible.

docs/quickstart.rst
nitishr: Thanks for the note, @fuhrysteve. I think you misunderstood a bit, so I've changed the example itself to use plain sqlalchemy instead of flask-sqlalchemy. For someone using flask-sqlalchemy, it should be pretty obvious they'd need to pass db.session instead of a 'hand-constructed' session.

fuhrysteve: Awesome, and yeah I agree that makes it a little more obvious.

mattupstate:
Flask-SQLAlchemy removes a lot of SQLAlchemy boilerplate that I personally know the majority of developers using Flask-Security appreciate. If you want plain-old SQLAlchemy, then there's nothing stopping you from implementing this on your own and providing Flask-Security with your own datastore.

nitishr:
Can't disagree with the benefits of Flask-SQLAlchemy, @mattupstate. And the good thing is, this change has no net-negative or net-positive impact on developers who already use Flask-SQLAlchemy (except for a really small avoidable incompatibility that I'll get to in a moment). It does, however, have a net positive effect for those who don't (want to) use Flask-SQLAlchemy (they won't need to implement their own datastore), and for Flask-Security itself (one less dependency).

Now, about the only difference this makes for existing Flask-Security users: SQLAlchemyUserDatastore needs to be initialized as:

SQLAlchemyUserDatastore(db.session, User, Role)

instead of:

user_datastore = SQLAlchemyUserDatastore(db, User, Role)

I have a fix for this in mind as well, which I'm happy to push as an update to this PR if you think this is useful.

janneraiskila:
This is an excellent idea.

The boilerplate caused by plain SQLAlchemy in a Flask app is really negligible compared to the separation benefits it gives. Thus a lot of people use plain SQLAlchemy. I think it would be beneficial to support it out of the box.

However, backwards compatibility for users of Flask-SQLAlchemy should definitely be maintained at the call site.

nitishr:
Just pushed a change to maintain call-site backwards compatibility. However, the Travis CI build seems to not have been triggered. Is that because "This branch has conflicts that must be resolved"?

docs/quickstart.rst
```diff
@@ -7,15 +7,15 @@ Quick Start
...
+============================
 
 SQLAlchemy Install requirements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ::
 
      $ mkvirtualenv <your-app-name>
-     $ pip install flask-security flask-sqlalchemy
+     $ pip install flask-security sqlalchemy
```
jirikuncar, line 18: We should keep the less chatty example with flask-sqlalchemy in the documentation and the "advance" one after.

tests/conftest.py
```diff
@@ -149,12 +145,53 @@ class User(db.Document, UserMixin):
 
 
 @pytest.fixture()
-def sqlalchemy_datastore(request, app, tmpdir):
-    from flask_sqlalchemy import SQLAlchemy
```
jirikuncar: We need to keep the `flask_sqlalchemy` in the tests. Could you add another fixture for the "pure" sqlalchemy version?

PR #429, opened 2015-10-14
Properly handle flask-login is_authenticated changes

jonafato:
As of version 0.3.0 of flask-login, the is_authenticated methods of
UserMixin and AnonymousUserMixin are properties instead of normal
methods. This change adds compatibility for both versions of these
methods.

Additionally, pytest-cache has been [merged into pytest](http://pytest.org/latest/changelog.html#id3) and is no longer
separately maintained, so pytest-cache-related requirements are being
updated.


jonafato:
As of the merging of #417, this is no longer relevant.

PR #431, opened 2015-10-16
Add option URL_SCHEME

houjunchen:
This option can be used to set url scheme to **https**.

I deployed a flask app which only accepts https requests and got a connection issue when I tried to reset password by the instruction sent to me since the link in that email was generated with scheme 'http'.

So I'm trying to fix this issue by this PR.


jonafato:
I think this is a problem better solved by [Flask-SSLify](https://github.com/kennethreitz/flask-sslify). Also see the Flask [docs on proxy setups](http://flask.pocoo.org/docs/0.10/deploying/wsgi-standalone/#proxy-setups), as they document required configuration for running applications behind proxies like nginx.

PR #437, opened 2015-11-03
Fix tests, add py35

jonafato:
- Add Python 3.5 support on travis

- pytest-cache has been merged into pytest [0] and is no longer
  separately maintained, so pytest-cache-related requirements are being
  updated.

- Flask-Mongoengine latest stable is broken on Python 2.6 (but fixed in
  master [1]), so its version is being restricted for now.

[0] http://pytest.org/latest/changelog.html#id3
[1] https://github.com/MongoEngine/flask-mongoengine/pull/174

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,20 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
   - "pypy"
 
 install:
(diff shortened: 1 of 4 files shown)


nfvs:
+1

Running `python setup.py test` on a clean environment fails since it installs the latest pytest version:
```setup.py: error: unrecognized arguments: --clearcache```

mattupstate:
LGTM

Issue #426, opened 2015-09-30
Add configuration setting SECURITY_BCRYPT_ROUNDS


vaughnkoch:
Or something similar. Passlib provides the option of specifying the # of rounds via the CryptContext object below, but that isn't available at the Flask-Security level because the call is hardcoded.

```
def _get_pwd_context(app):
    pw_hash = cv('PASSWORD_HASH', app=app)
    schemes = cv('PASSWORD_SCHEMES', app=app)
    deprecated = cv('DEPRECATED_PASSWORD_SCHEMES', app=app)
    if pw_hash not in schemes:
        allowed = (', '.join(schemes[:-1]) + ' and ' + schemes[-1])
        raise ValueError("Invalid hash scheme %r. Allowed values are %s" % (pw_hash, allowed))
    return CryptContext(schemes=schemes, default=pw_hash, deprecated=deprecated)
```

It would be nice to have a setting that could specify the # of rounds. This could be useful in unit testing for example, when you still want to exercise bcrypt but want to use 1 or 2 rounds instead of 12.

PR #441, opened 2015-11-10
Add support for a SECURITY_BCRYPT_ROUNDS setting.

nfvs:
Allows the usage of a variable number of rounds when using bcrypt. Set
the rounds to 4 in test_hashing.py, and add a test for an invalid number of
rounds (e.g. 3).
#426

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -53,6 +53,8 @@ Core
 ``SECURITY_DEPRECATED_HASHING_SCHEMES``  List of deprecated algorithms used for
                                          creating and validating tokens.
                                          Defaults to ``hex_md5``.
+``SECURITY_PASSWORD_HASH_OPTIONS``       Specifies additional options to be passed
+                                         to the hashing method.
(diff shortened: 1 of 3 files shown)


nfvs:
bump?

nfvs:
@jonafato any chance of merging this? It's a pretty useful security feature and seems like a harmless change.

jirikuncar:
@nfvs can we go with some more generic solution which would allow people to define keyword arguments for any hashing method?

```python
SECURITY_PASSWORD_HASH_OPTIONS = {
    'bcrypt': {'rounds': 4},
}

_pwd_context.encrypt(
    signed, 
    **config_value('PASSWORD_HASH_OPTIONS').get(_security.password_hash, {})
)
```

Alternatively you can register new hash function with your defaults: http://passlib.readthedocs.io/en/stable/lib/passlib.registry.html?highlight=register_crypt_handler_path#passlib.registry.register_crypt_handler_path.

nfvs:
@jirikuncar great idea. I've modified it to implement that instead.

PR #442, opened 2015-11-12
Migrate to Travis container-based infrastructure, optimize build with pip caching.

nfvs:
Migrate to Travis container-based infrastructure.

Advantages are listed here:
http://docs.travis-ci.com/user/migrating-from-legacy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ python:
 
 sudo: false
 
+cache:
+  - apt
+  - pip
+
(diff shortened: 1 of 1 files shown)


nfvs:
Smaller gain than I had hoped for (~3mins) in total run time of the tests, but still worth it IMO, especially since the old infrastructure is now marked as legacy.

jonafato:
Awesome, thanks!

PR #444, opened 2015-11-18
DISABLE_ENCRYPTION option

charliewolf:
I use the wonderful [SQLAlchemy-Utils](http://sqlalchemy-utils.readthedocs.org/en/latest/data_types.html) library which transparently handles password encryption in your database models. However, because this is not a String, but rather a PasswordType it is not compatible with Flask-Security's encryption engine, even when set on 'plaintext'. This gives the option to disable encryption entirely.

### DIFF
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -131,7 +131,7 @@ def verify_and_update_password(password, user):
     :param user: The user to verify against
     """
     if config_value('DISABLE_ENCRYPTION'):
-        return password == user.password        
+        return password == user.password
     if _pwd_context.identify(user.password) != 'plaintext':
         password = get_hmac(password)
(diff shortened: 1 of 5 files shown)


flask_security/utils.py
```diff
@@ -130,7 +130,8 @@ def verify_and_update_password(password, user):
     :param password: A plaintext password to verify
     :param user: The user to verify against
     """
-
+    if config_value('DISABLE_ENCRYPTION'):
+        return password == user.password
```
jirikuncar: please remove trailing whitespace so Travis is green ;)

jirikuncar:
@charliewolf thanks for this PR. I was about to propose the same. Good job!

charliewolf:
@jirikuncar Good catch. Fixed.

PR #445, opened 2015-11-19
Set Flask Global Context User in _core _user_loader() function

gxela:
sorry no tests, I looked but couldn't find a place to test if global context user gets set on login. fixes things on my side where global context user is in use.


jonafato:
Does this provide any functionality that is not already addressed by Flask-Login's [`current_user`](https://flask-login.readthedocs.org/en/latest/#flask.ext.login.current_user)?

PR #449, opened 2015-11-25
utils: use `None` instead of 'untrackable' for IPs

### LINKED ISSUE #448: utils: use `None` instead of 'untrackable' when IP is not detected
### Problem

Using `IPAddressType` is not possible when used together with string value `'untrackable'`.

``` python
from sqlalchemy_utils import IPAddressType
# ...
class User(db.Model):
    # ...
    last_login_ip = db.Column(IPAddressType, nullable=True)
    current_login_ip = db.Column(IPAddressType, nullable=True)
    # ...
    @validates('last_login_ip', 'current_login_ip')
    def validate_ip(self, key, value):
        """Hack untrackable IP addresses."""
        if value == 'untrackable':
            value = None
        return value
```
### Proposal

Store `None` value instead of `'untrackable'` in [`flask_security.utils:login_user`](https://github.com/mattupstate/flask-security/blob/4049c0620383f42d37950c7a35af5ddd6df0540f/flask_security/utils.py#L68) when the IP address of user can not be detected.

``` diff
diff --git a/AUTHORS b/AUTHORS
index 1aaa781..1d3c452 100644
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,6 +22,7 @@ Eric Butler
 Eskil Heyn Olsen
 Iuri de Silvio
 Jay Goel
+Jiri Kuncar
 Joe Esposito
 Joe Hand
 Josh Purvis
diff --git a/flask_security/utils.py b/flask_security/utils.py
index ccd5b0b..114421c 100644
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -65,7 +65,7 @@ def login_user(user, remember=None):
         if 'X-Forwarded-For' in request.headers:
             remote_addr = request.headers.getlist("X-Forwarded-For")[0].rpartition(' ')[-1]
         else:
-            remote_addr = request.remote_addr or 'untrackable'
+            remote_addr = request.remote_addr or None  # make sure it is None and not ''

         old_current_login, new_current_login = user.current_login_at, datetime.utcnow()
         old_current_ip, new_current_ip = user.current_login_ip, remote_addr
diff --git a/tests/test_trackable.py b/tests/test_trackable.py
index 7017754..a0c8906 100644
--- a/tests/test_trackable.py
+++ b/tests/test_trackable.py
@@ -23,7 +23,7 @@ def test_trackable_flag(app, client):
         user = app.security.datastore.find_user(email=e)
         assert user.last_login_at is not None
         assert user.current_login_at is not None
-        assert user.last_login_ip == 'untrackable'
+        assert user.last_login_ip is None
         assert user.current_login_ip == '127.0.0.1'
         assert user.login_count == 2

@@ -39,6 +39,6 @@ def test_trackable_with_multiple_ips_in_headers(app, client):
         user = app.security.datastore.find_user(email=e)
         assert user.last_login_at is not None
         assert user.current_login_at is not None
-        assert user.last_login_ip == 'untrackable'
+        assert user.last_login_ip is None
         assert user.current_login_ip == '88.88.88.88'
         assert user.login_count == 2
```

jirikuncar:
Changes trackable behavior of `login_user` when IP can not be detected
from a request from 'untrackable' to `None` value.  (closes #448)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,6 +22,7 @@ Eric Butler
 Eskil Heyn Olsen
 Iuri de Silvio
 Jay Goel
+Jiri Kuncar
 Joe Esposito
 Joe Hand
 Josh Purvis
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -65,14 +65,14 @@ def login_user(user, remember=None):
         if 'X-Forwarded-For' in request.headers:
             remote_addr = request.headers.getlist("X-Forwarded-For")[0].rpartition(' ')[-1]
         else:
-            remote_addr = request.remote_addr or 'untrackable'
+            remote_addr = request.remote_addr or None  # make sure it is None and not ''
 
         old_current_login, new_current_login = user.current_login_at, datetime.utcnow()
         old_current_ip, new_current_ip = user.current_login_ip, remote_addr
 
         user.last_login_at = old_current_login or new_current_login
         user.current_login_at = new_current_login
-        user.last_login_ip = old_current_ip or new_current_ip
+        user.last_login_ip = old_current_ip
         user.current_login_ip = new_current_ip
         user.login_count = user.login_count + 1 if user.login_count else 1
 
--- a/tests/test_trackable.py
+++ b/tests/test_trackable.py
@@ -23,7 +23,7 @@ def test_trackable_flag(app, client):
         user = app.security.datastore.find_user(email=e)
         assert user.last_login_at is not None
         assert user.current_login_at is not None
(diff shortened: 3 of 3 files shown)


jirikuncar:
@mattupstate would you consider following change? We could then avoid having our custom IP address validator (see https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L81-L89). Thanks

jirikuncar:
@mattupstate any chance that you might have a look after a year?

Issue #421, opened 2015-09-22
REST authentication ?


asmodehn:
Is there a way that I can implement a REST API to be able to authenticate users using flask security ?

From what I could gather from the documentation around, flask-login and flask-security seems to aim at securing websites, however I need to build a REST authentication service ( and an "internal API" for other services to check that users are authenticated ) and I would obviously like to avoid reimplementing everything...

How/where should I plug my API to reuse a maximum of flask-security while not using the templates ? By any chance, is there any project already doing this that you are aware of ?

=> It seems we could have two main parts in flask-security :
- the templates and mechanisms for login/password & sessions handling that people use for securing website.
- the backend integration and authentication/authorization mechanism that can also be used by people wanting to implement authentication REST services.

Thanks for letting me know how you would go about doing this, I ll probably start on it sometime soon...

fuhrysteve:
So are you looking for a sort of stateless authentication with flask security? Check out `login_manager.request_loader`.

Here's how you might implement this functionality using `flask-security`

```python
user_datastore = SQLAlchemyUserDatastore(db, User, Role)
security = Security(app, user_datastore)

@security.login_manager.request_loader
def load_user_from_request(req):
    api_key = req.args.get('api_key', req.headers.get('Authorization'))
    # retrieve and return a user object
```

davvid:
I'm using flask-security and I find that I am able to authenticate using flask-security by setting the content-type to application/json and issuing regular json requests.

The only difficulty you might encounter is the need to provide csrf tokens for flask-wtf.  The changes in my PR makes that possible.

Flask-Security expects an `Authentication-Token` header with the value from the user token provided from the login endpoint when using its `auth_required` decorator.  Post credentials to the login endpoint and you'll get back the user's token.  That token can be used to access views protected with `auth_required`.

asmodehn:
Thanks everyone for the useful feedback.
I haven't been able to start working on it yet, but I thought I should explain more what I am trying to do here...

I am currently in the process of building a set of microservices, REST based, one service for one purpose. I haven't started the authentication / authorization part yet.

An Authentication microservice could be quite generic and could be already out there somewhere ( how about https://github.com/LukeB42/microauth ? ).
But to verify that a user is authenticated in all other services, there must be some non-trivial flow of information exchange ( not only "check the DB" but more like "move around the token to the proper service at the proper time" ), and there are multiple possible solution.

Ideally, I would love to find : 
1) a (flask based) microservice that only cares about authentication
[...]

PR #460, opened 2015-12-28
views: allow rendering json forms when form.user is unset

davvid:
Doing a GET against /security/login with content-type: application/json
results in a traceback because form.user is accessed sans checking for
existence.

Make the request succeed by checking for the user before accessing its
fields.  A subsequent commit will make this useful in other situations.

Related-to: #421


jirikuncar:
All authenticated requests using Flask-Security need a user.

Issue #30, opened 2012-08-10
Allow manual authentication


leaanthony:
Hi, I'm currently using the stable version and am finding it really hard to be able to authenticate users manually. The main reason for requiring this is to get authentication working via an ajax call. I've noticed in the dev branch there is a utility function that looks like it might do this. Is this my only option?

Cheers.

mandarvaze:
@melpomene 
Setting `WTF_CSRF_ENABLED=False` in your application's configuration gets rid of `CSRF Token missing` error.

bschumacher:
@mandarvaze @melpomene 

Disabling CSRF isn't really a good idea. You can get the current token value like this:

from flask import jsonify
from flask.ext.security import LoginForm

@app.route('/login_csrf')
def login_csrf():
    form = LoginForm()
    return jsonify({'csrf_token': form.csrf_token.current_token})

You can then use the CSRF token that is returned as JSON in your login attempt, simply pass it as csrf_token with your email and password.

radusuciu:
@bschumacher Did you attempt your suggested solution?

`requests.post(url + '/login', 
    data=json.dumps({ 
        'email': email, 
        'password': password,
        'csrf_token': requests.get(url + '/login_csrf').json()['csrf_token'] 
    }),
    headers={'content-type':'application/json'}
).json()`

Response:

`{'meta': {'code': 400}, 'response': {'errors': {'csrf_token': ['CSRF token missing']}}}`

radusuciu:
If like me you are not married to getting that auth token then you may follow this route using cookies instead:

```python
csrf_req = requests.get(url + '/login_csrf')
csrf_token = csrf_req.json()['csrf_token']

login_req = requests.post(url + '/login',
    data={ 
        'email': email, 
        'password': password,
        'csrf_token': csrf_token
    },
    cookies=csrf_req.cookies,
    allow_redirects=False
)
```

Finally, you can then do:

```python
cookies = login_req.cookies
requests.get(url, cookies=cookies)
```

You may use the cookies for further requests. If you don't want to setup that `/login_csrf` route you can use `BeautifulSoup` too:

```python
csrf_req = requests.get(url + '/login', allow_redirects=False)
csrf_token = BeautifulSoup(csrf_req.text).find('input', id='csrf_token')['value']
```

and then proceed as above.

spitzbubchen:
@radusuciu Any ideas on how to logout and destroy your token cookie?

radusuciu:
[...]

PR #461, opened 2015-12-28
views: allow adding a csrf token to json responses

davvid:
flask-wtf's CsrfProtect makes it hard to use flask-security with json
requests because a csrf token is needed when posting json data.

This difficulty can be alleviated by teaching flask-security
provide a csrf token in its responses.

Teach _render_json() to supply a csrf_token value in the response when
the SECURITY_JSON_INCLUDE_CSRF configuration is True.  The presence
of csrf_token allows clients to respond back with the X-CSRFToken
header set to its value.  This allows flask-security json
requests to work without needing to disable csrf.

Related-to: #421

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -47,6 +47,13 @@ Core
 ``SECURITY_DEFAULT_HTTP_AUTH_REALM``     Specifies the default authentication
                                          realm when using basic HTTP auth.
                                          Defaults to ``Login Required``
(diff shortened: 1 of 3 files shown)


radusuciu:
This may help with #30?

spitzbubchen:
Just wondering if this has been implemented? If so can an example be provided?

jirikuncar:
I would recommend to use OAuth instead.

PR #462, opened 2015-12-28
views: allow token authentication for change_password

davvid:
Allow token authentication for the change_password endpoint.
This makes it possible change passwords from a json request.

Related-to: #421


jirikuncar:
Cherry-picked to #785

PR #465, opened 2016-01-02
core: fix infinite recursion in __getattr__

davvid:
The intention of **getattr** is to return the value from the internal
state.  When using the factory pattern, the instance's _state is never
set, so __getattr__ fails:

```
flask_security/core.py:453: in __getattr__
    return getattr(self._state, name, None)
!!! Recursion detected (same locals & position)
```

We can make both use cases work by providing a default _state that
points to the security local proxy at the class level.  This allows the
immediate-constructor Security(app) form to override the class attribute
in the instance, as it currently does, while also allowing
__getattr__ to succeed when using the factory pattern.

Add tests to exercise both the existing and newly-fixed behaviors.

Related-to: #317

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -280,3 +280,15 @@ def test_custom_forms_via_config(app, sqlalchemy_datastore):
 def test_without_babel(client):
     response = client.get('/login')
     assert b'Login' in response.data
+
+
+def test_getattr_with_app_constructor(app, sqlalchemy_datastore):
+    security = Security(app=app, datastore=sqlalchemy_datastore)
+    assert security.flash_messages is True
+
+
+def test_getattr_with_default_constructor(app, sqlalchemy_datastore):
+    security = Security(datastore=sqlalchemy_datastore)
+    security.init_app(app)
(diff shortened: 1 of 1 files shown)


jxltom:
This one-line solution looks good to me. But it is not merged or has any feedback after a year... The```core.py``` has conficts now.

jirikuncar:
@davvid it looks like an elegant solution, however there were some changes meantime. Can you please rebase and check that it is still working? Thanks!

davvid:
Thanks, I rebased and verified that the tests still  pass.

flask_security/core.py
```diff
@@ -493,6 +493,8 @@ class Security(object):
     :param anonymous_user: class to use for anonymous user
     """
 
+    _state = _security
```
jirikuncar: @davvid have you checked that it is really needed after the rebase?

davvid:
I removed the class attribute and everything works great.  These tests previously failed.  The commit has been amended in case you want to take just the tests.

`init_app()` now sets `self._state`, so this PR can be closed otherwise.  cheers

Issue #453, opened 2015-12-01
reset_password for restful API


yingbo:
The forgot_password in view.py supports  it, but not reset_password(token). Would you provide the same function for reset_password(token)? I tried on my local and looks followings work:

```python
    #form = _security.reset_password_form()    
    form_class = _security.reset_password_form

    if request.json:
        form = form_class(MultiDict(request.json))
    else:
        form = form_class()

```

alaa-aqeel:
i tried this **to make `reset_password` for restful API**   and it work 
```
def reset_password(token):
     expired, invalid, user = reset_password_token_status(token)

    if invalid:
        if request.is_json:
            return _render_json_msg('invalid_token','INVALID_RESET_PASSWORD_TOKEN') , 400 
        do_flash(*get_message('INVALID_RESET_PASSWORD_TOKEN'))
    if expired:
        send_reset_password_instructions(user)
        if request.is_json:
            return _render_json_msg('expired_token','PASSWORD_RESET_EXPIRED') ,400
        do_flash(*get_message('PASSWORD_RESET_EXPIRED', email=user.email,
                              within=_security.reset_password_within))
    if invalid or expired:
        return redirect(url_for('forgot_password'))

    form = _security.reset_password_form()

    if form.validate_on_submit():
        after_this_request(_commit)
        update_password(user, form.password.data)
        # login_user(user)
        #testing
        if not request.is_json:
            do_flash(*get_message('PASSWORD_RESET'))
            login_user(user)
            return redirect(get_url(_security.post_reset_view) or
                        get_url(_security.post_login_view))
    #resting
    if request.is_json:
        form.user = current_user
        return _render_json(form)

    return _security.render_template(
        config_value('RESET_PASSWORD_TEMPLATE'),
        reset_password_form=form,
        reset_password_token=token,
        **_ctx('reset_password')
    )
```
**_render_json_msg**
```
def _render_json_msg(key,msg_key):
    response = dict(errors={})
    response['errors'][name] = [*get_message(msg_key)]
    return jsonify(response=response)
```

jwag956:
Please consider trying out:
https://pypi.org/project/Flask-Security-Too/#history

release 3.2.0rc3
This has support for JSON for all endpoints, including support for single-page-applications that need non-form redirects. It also has a swagger/openapi specification for the current API.

It would be great if this worked for you...

alaa-aqeel:
@jwag956  **nice**

PR #468, opened 2016-01-10
The change makes reset password from API.

yingbo:
It makes reset_password(token) supporting call from API besides web forms.
Related to #453


yingbo:
@mattupstate , any thoughts on this pull request?

Issue #340, opened 2014-11-14
Maximum recursion depth exceeded 


eriktaubeneck:
If the `security` object gets initialized before the `app` exists (i.e. using the factory method) and `init_app` is called later, `security._state` is never set. This makes sense because `_state` is dependent upon the `app`, however, when an attribute on `security` is accessed, the redefinition of `__getattr__` looks for `_state`, which isn't set, so `__getattr__` get's called, looks for `_state`, etc, and then bam

```
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

I opened a [PR](https://github.com/mattupstate/flask-security/pull/339), but closed it because I'm not sure if it's the right fix. I solved the issue myself by doing:

```
security = app.extensions['security']
@security.login_context_processor
...
```

but at the very least, I think there should be a better exception that gets raised (the max recursion depth really threw me off for a bit.) The solution in the PR would work, but would require the above to become something like

```
with app.app_context():
    @security.login_context_processor
    ...
```

which maybe works? Curious what other people think is a good solution.

bbenne10:
Is there a reason that you simply couldn't do `self._state = state` just before `return state` in core.py's `init_app`? I've got a fork that does just that, and it seems to resolve this issue (though not the one I was trying to resolve...)

mcs07:
I get this recursion problem when using `@security.send_mail_task` after doing a `security.init_app()` in my Flask application factory. @bbenne10's `self._state = state` solution seems to fix it for me.

eriktaubeneck:
@bbenne10 @mcs07 the issue with setting `self._state = state` is that `security.init_app` is designed to be used with the [Flask factory pattern](http://flask.pocoo.org/docs/0.10/patterns/appfactories/), and an instance of `security` needs to be able to work with multiple instances of a Flask `app`. `app.extensions['security']` returns the `state` which is associated with the specific instance of `app`, which is what allows my above solution to work.

nfvs:
@bbenne10 I have a pull request pending #255 that does exactly that (besides other things), not sure when/if it will be merged though.

eriktaubeneck:
@nfvs it won't be. see [this comment](https://github.com/mattupstate/flask-security/issues/211) from last year. the fact that `self._state` doesn't get set to `state` in the factory method is intentional.

nfvs:
@eriktaubeneck awesome lookup skills, I have modified my PR accordingly :)

eriktaubeneck:
[...]

PR #471, opened 2016-01-21
fix security.init_app(app) usage by setting self._state

wassname:
Bugfix: `security.init_app(app)` doesn't match `security = Security(app)` as the first doesn't set security._state while the second does. This change fixes that.

Without this change `security.send_mail_task` will give an infinite recursion error.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -423,7 +423,7 @@ class Security(object):
 
         identity_loaded.connect_via(app)(_on_identity_loaded)
 
-        state = _get_state(app, datastore,
+        self._state = _get_state(app, datastore,
                            login_form=login_form,
                            confirm_register_form=confirm_register_form,
                            register_form=register_form,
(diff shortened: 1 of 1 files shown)


flask_security/core.py
```diff
@@ -435,11 +435,11 @@ def init_app(self, app, datastore=None, register_blueprint=True,
...
         if register_blueprint:
-            app.register_blueprint(create_blueprint(state, __name__))
+            app.register_blueprint(create_blueprint(self._state, __name__))
             app.context_processor(_context_processor)
 
         state.render_template = self.render_template
-        app.extensions['security'] = state
+        app.extensions['security'] = self._state
 
         return state
```
jxltom: ```state.render_template = self.render_template``` should be ```self.state.render_template = self.render_template```

jxltom:
After searched for a while, I found this may not be a bug or this may not be the right way to fix. I have some explain [here](https://github.com/mattupstate/flask-security/issues/340#issuecomment-278560428).

PR #479, opened 2016-02-04
Enable token authentication for change_password()

tomazberisa:
Two changes:
1. Enable changing password by providing token
2. Provide new token in response to successful password change


jirikuncar:
moved to #785

Issue #466, opened 2016-01-04
datetime objects used are not timezone aware


jadkik:
The `datetime.utcnow()` function is used anywhere the current datetime is needed. That function returns an offset-naive datetime object by design in [Python 2](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcnow) and [Python 3](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow).

This causes problems when using the `arrow` library for dates and times using e.g. SQLAlchemy with [`sqlalchemy_utils`](http://sqlalchemy-utils.readthedocs.org/en/latest/data_types.html#sqlalchemy_utils.types.arrow.ArrowType) installed.

Logging in after the first time fails because for some reason the dates are compared by sqlalchemy before committing the transaction, resulting in the following exception:

```
TypeError: can't compare offset-naive and offset-aware datetimes
```

If the datetime objects were timezone aware, there would not be a problem. And I think some database backends do store this information, so it might be better in those cases too. (I'm using MySQL, so it does not affect me in this regard)

avilaton:
I hit the same issue. We are using postgresql and our tracking fields are created as `TIMESTAMP WITH TIME ZONE` and the user model uses `db.Column(db.DateTime(timezone=True))`. On the second login we get an error
```
Traceback (most recent call last):
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/gavila/.virtualenvs/st/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
[...]

PR #483, opened 2016-02-12
Add timezone to datetime.utcnow() in current_login

avilaton:
Proposal for closing #466. Please @jadkik check this out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,19 @@ cache:
   - pip
 
 install:
+  - |
+      if [ "$TRAVIS_PYTHON_VERSION" = "pypy" ]; then
+        export PYENV_ROOT="$HOME/.pyenv"
+        if [ -f "$PYENV_ROOT/bin/pyenv" ]; then
+          cd "$PYENV_ROOT" && git pull
+        else
+          rm -rf "$PYENV_ROOT" && git clone --depth 1 https://github.com/yyuu/pyenv.git "$PYENV_ROOT"
+        fi
+        export PYPY_VERSION="4.0.1"
+        "$PYENV_ROOT/bin/pyenv" install "pypy-$PYPY_VERSION"
+        virtualenv --python="$PYENV_ROOT/versions/pypy-$PYPY_VERSION/bin/python" "$HOME/virtualenvs/pypy-$PYPY_VERSION"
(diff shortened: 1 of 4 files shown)


avilaton:
How naive was I, this PR fails... The current sqlalchemy tests do not use timezone so comparing them fails. Maybe it has to be made into an config option to use timezones, @mattupstate probably has a better idea.

jonafato:
Thanks for the PR! It seems like there's demand for this feature, so it's definitely worth considering. A few notes:

1. The solution should be configurable and default to the current behavior to preserve backward compatibility.

2. `pytz` is a bit of a heavy dependency to add just for this purpose. Python 3.2+ includes [`datetime.timezone.utc`](https://docs.python.org/3/library/datetime.html#datetime.timezone.utc). The Python 2 [`tzinfo` documentation](https://docs.python.org/2/library/datetime.html#tzinfo-objects) includes an example UTC tzinfo class definition. It seems like using these (possibly in the 2-to-3 compatibility sections of `utils.py`) would be sufficient to fix the issue.

3. If a third party solution is used, its requirement should not be pinned to a specific version. With the case of `pytz` (and any other library that tracks timezone information), you likely always want the most up-to-date version.

4. `pytz` includes all (500+) of the Olson timezone files in its distribution. This is unfortunately a problem on Google AppEngine (which has a very small file limit), leading to libraries like [gaerpytz](https://gaerpytz.appspot.com/) as workarounds. Adding pytz as a Flask-Security dependency would break applications being deployed on GAE, which is undesirable.

TL;DR: the solution here is probably configurable behavior to use a UTC `tzinfo` implementation taken directly from the Python docs rather than depending on another library.

avilaton:
Thanks to you @jonafato for the explanation. I guess I would have to write a test suite that uses something other than sqlite to be able to test the feature. I will try to figure out how to do that.

erwagasore:
Is it possible to make this `new_current_login_at` configurable similarly to `SECURITY_TRACKABLE` and let user choose what to do being naive or aware

```python
SECURITY_TRACKABLE = True
# SECURITY_DATE_AWARENESS = datetime.utcnow()
SECURITY_DATE_AWARENESS = arrow.utcnow()
```

This will arrow users to choose which library to pip install

philsalesses:
+1 This is a critical bug imho

PR #484, opened 2016-02-12
Avoid exception when posting empty email or password

tomazberisa:
Previously, when posting empty email or password, the following exception was raised:
AttributeError: 'NoneType' object has no attribute 'strip'


jirikuncar:
Closed via 292f89c2

PR #485, opened 2016-02-15
Allow non-email check_http_auth

pnpnpn:
Should be using `get_user` instead of `find_user`. The `get_user` is more flexible in that it does not require username to be an email.


jirikuncar:
@pnpnpn can you please rebase the PR and include a test? Thanks!

PR #487, opened 2016-02-19
Use Security.render_template in mails too

noirbizarre:
This PR use `Security.render_template` in mails too (allow custom `render_template` function, mostly for theming support)


noirbizarre:
Sorry @jirikuncar, I just saw the approval. It's rebased and conflict are resolved

PR #495, opened 2016-03-26
[WIP] Fixes #340 factory pattern support

### LINKED ISSUE #340: Maximum recursion depth exceeded

KeNaCo:


flask_security/core.py
```diff
@@ -26,6 +30,9 @@
     ForgotPasswordForm, ChangePasswordForm, ResetPasswordForm, \
     SendConfirmationForm, PasswordlessLoginForm
 
+# Select stack, in flask 0.9 _app_ctx_stack is used
```
jxltom: Maybe this will be more clear?

```python
try:
    from flask import _app_ctx_stack as _ctx_stack
except ImportError:
    from flask import _request_ctx_stack as _ctx_stack
```

flask_security/core.py
```diff
@@ -423,28 +440,31 @@ def init_app(self, app, datastore=None, register_blueprint=True,
...
         return render_template(*args, **kwargs)
 
     def __getattr__(self, name):
-        return getattr(self._state, name, None)
+        if self._state is not None:  # shortcut
+            return getattr(self._state, name)
+        ctx = _ctx_stack.top
+        if ctx is not None:
+            self._state = get_state(ctx.app)
+            return getattr(self._state, name)
```
jxltom: Will code below ```ctx = _ctx_stack.top``` ever run? Since ```self._state``` is alway being set in ```__init__``` or ```init_app```, maybe these are additional?

jxltom:
Actually this PR also has another commit which try to fix inconsistent datastore between ```__init___``` and ```init_app```. But as author said [here](https://github.com/mattupstate/flask-security/pull/537#issuecomment-272888626) in #537, this is not an intended API, so it will be better to pull request separately and discuss what is the proper way for initialize datastore in another PR.

PR #500, opened 2016-04-19
Add config option to skip sending password reset email

mafrosis:
My app hooks into the Flask Security signals to send email via a different method. This change means the password reset emails can be disabled to achieve this.

Second commit also passes the token to the `confirm_instructions` signal.


mafrosis:
Closing as these commits made it into `develop` via the fork

PR #507, opened 2016-05-05
Fix import which caused pyflakes breakage when running tests

tomers:


jonafato:
This appears to be related to https://github.com/pyflakes/pyflakes/pull/61. Once that fix is merged (or some other fixing the same issue), the tests should return to their normal state.

jonafato:
pyflakes has published a bugfix release addressing this issue.. Builds are back to normal, so I'm going to close this pull request. Thanks for reporting it and providing a workaround.

PR #512, opened 2016-05-08
Fix #383

### LINKED ISSUE #383: Inconsistent behavior between Security(...) and init_app(...)

bkircher:
Fix inconsistency between Security and init_app


jirikuncar:
Closing as stalled. Many issues have been fixed in c657f32c.

Issue #513, opened 2016-05-08
Default login/register forms accept wildcards for email address


AshtonShapcott:
When using the [SQLAlchemy quick start](https://pythonhosted.org/Flask-Security/quickstart.html#id1) code this is easy to replicate by specifying your email address as `%` during login. This also applies to registration.

This seems to be caused by `SQLAlchemyUserDatastore.get_user(...)` accepting wildcards.

I am not certain if wildcard matching is an intended behavior in `get_user` given it only returns a single value and is likely to create conflicts with user identifiers. An example using an email identifier: `user@example.com` and `us_r@example.com` cannot both be registered in the same datastore even though they're unique emails potentially held by different users.

radusuciu:
This seems like it might help: https://github.com/mattupstate/flask-security/pull/516

jirikuncar:
Closed via #516

PR #514, opened 2016-05-08
Fix SQLAlchemy get_user(..) allowing wildcards.

AshtonShapcott:
Potential fix for #513 under the assumption wildcards aren't an intended feature of `SQLAlchemyUserDatastore.get_user(...)`.


jirikuncar:
Solved in #516

PR #516, opened 2016-05-09
Do exact case insensitive match when looking up user

jasco:
Prevents wildcard matches supported ilike


jasco:
Updated.  Rebased to head and extended patch to support peewee db.

Issue #496, opened 2016-04-12
Two Factor Authentication


Galstat:
Hi,
I'm currently working on an application which uses flask-security mechanisms. 
I was looking to add a second factor authentication to my application and couldn't find a feature inside that can be used for it. Do you guys know any external feature or plugin that can be integrated to it? 

Will you be willing to accept a pull request on that?

Thanks,
Gal

Galstat:
I have created a pull request #522 
for this issue, stay safe!

pvsune:
your PR hasn't been touched in a while. i'm looking into adding 2fa too in my app. would you have time to look at it a bit? would gladly help fill out missing parts if any.

LoganHenderson:
Very interested in using this for a current project of mine. Is there anything I can do to help move this forward? It seemed like the PR was failing the CI tests from just merge conflicts

PR #522, opened 2016-06-01
Security two factor authentication feature

Galstat:
* supporting two factor authentication:

* Support Mail, SMS, or Google Authenticator second factor authentication.

* Ability to change second factor authentication to existing users 

* Provide rescue mail in case of lost phone

* update docs with two factor authentication changes

* updating requirements, authors

* adding two_factor test

* improving tests coverage

Related to issue #496 

https://github.com/mattupstate/flask-security/issues/496

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ Session.vim
 *~
 
 .eggs/README.txt
+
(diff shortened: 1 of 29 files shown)


maxnordlund:
This looks lovely, but I guess a bit lost to the sands of time. Would you mind rebasing @Galstat? Then let's hope the @mattupstate can take a look.

PR #528, opened 2016-07-07
Use new style flask imports

PlasmaSheep:
In the most recent versions of flask, importing `flask.ext.*` will result in a warning message. The preferred way to import is `flask_*`. This PR updates the quickstart documentation to reflect this change.


jonafato:
Thanks!

PR #529, opened 2016-07-08
Py3 fix for getting user by email

cybertoast:
The exception thrown by MongoEngine in Py3 when checking a user's id is TypeError, vs ValidationError in Py2. This fix adds the check for TypeError.


jirikuncar:
Duplicate of 32bfd4e1399c9b53312c494bf6138b576a6251ca

PR #537, opened 2016-08-04
Fix init_app not setting datastore

abendebury:


PlasmaSheep:
Fixes issue #383

mattupstate:
This is not the API I intended. The `init_app` method is meant to serve the factory which, hypothetically, if you had multiple apps to initialize with the same datastore you could do so like:

```
security = Securit(datastore=yourdatastore)

app = Flask()
security.init_app(app)

another_app = app = Flask()
security.init_app(another_app)

abendebury:
So are you saying that #383 is not a bug or that this is not how the bug should be fixed?

PR #352, opened 2014-12-28
X-Forwarded-For can contain multiple IP addresses

fuhrysteve:
We should be using whatever the last IP in `X-Forwarded-For` is. Note that in order for this to work properly behind a trusted proxy (i.e. if you are using nginx / similar to forward to `gunicorn` / similar, as is common), you must use `ProxyFix` as described in the `werkzeug` and `flask` documentation. I've added a note to the docs for `SECURITY_TRACKABLE` to point that out, since it's probably not obvious. Perhaps a note should be added elsewhere too? Not sure.

From the nginx docs: http://nginx.org/en/docs/http/ngx_http_proxy_module.html

```
$proxy_add_x_forwarded_for
    the “X-Forwarded-For” client request header field with the $remote_addr variable
    appended to it, separated by a comma. If the “X-Forwarded-For” field is not present
    in the client request header, the $proxy_add_x_forwarded_for variable is equal to
    the $remote_addr variable.
```

Here's a stack trace illustrating the issue I was having while using ProxyFix behind a proxy, which in this case was `127.0.0.1`. The client's actual IP address in this case should be `189.254.205.210` (a Mexican IP address - the client was actually in Mexico at the time), though they were being proxied by an ec2 host at `172.21.5.123` (some indent added for clarity):

```python
DataError: (raised as a result of Query-invoked autoflush; consider using a 
session.no_autoflush block if this flush is occurring prematurely) (DataError)
invalid input syntax for type inet: "172.21.5.123, 189.254.205.210"
LINE 1: ...-27T00:00:25.154208'::timestamp, current_login_ip='172.21.5....
                                                             ^
 'UPDATE user_account 
SET
last_login_at=%(last_login_at)s,
current_login_at=%(current_login_at)s,
current_login_ip=%(current_login_ip)s,
login_count=%(login_count)s, modifieddate=now()
WHERE
user_account.user_id = %(user_account_user_id)s' {
    'current_login_at': datetime.datetime(2014, 12, 27, 0, 0, 25, 154208),
    'login_count': 14,
    'user_account_user_id': 3,
    'current_login_ip': '172.21.5.123, 189.254.205.210',
    'last_login_at': datetime.datetime(
        2014, 12, 26, 21, 21, 21, 456305,
        tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None))
}

Stacktrace (most recent call last):

  File "flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
[...]

PR #542, opened 2016-08-17
Use ProxyFix instead of inspecting X-Forwarded-For header

lnielsen:
PR #352 fixed a security issue in extracting a correct remote IP address from `X-Forwarded-For` header. 

The fix however doesn't take multiple trusted proxies into account, which you typically have in a load balanced setup (e.g. HAProxy + Nginx). The problem is located [here](https://github.com/mattupstate/flask-security/blob/develop/flask_security/utils.py#L65-L66), and can quickly be demonstrated with this example where it's always the last address in `X-Forwarded-For` being chosen:

``` python
>>> x_forwarded_for = '33.33.33.33, 22.22.22.22, 11.11.11.11'
>>> x_forwarded_for.rpartition(' ')[-1]
'11.11.11.11'
```

The problem with the current code is that if you use [ProxyFix](http://werkzeug.pocoo.org/docs/0.11/contrib/fixers/#werkzeug.contrib.fixers.ProxyFix) to set a correct IP address in `request.remote_addr`, then Flask-Security doesn't pick it up because X-Forwarded-For headers are inspected first. The workaround so far, is in addition to ProxyFix to use [HeaderRewriterFix](http://werkzeug.pocoo.org/docs/0.11/contrib/fixers/#werkzeug.contrib.fixers.HeaderRewriterFix) to remove the X-Forwarded-For headers completely from the request.
### Proposal

IMHO, Flask-Security (and any other Flask extension for that matter) should rely solely on `request.remote_addr` to get the client IP address, and only document that you should use ProxyFix in order to ensure that `remote_addr` is set correctly ([already done in Flask-Security](https://pythonhosted.org/Flask-Security/configuration.html#feature-flags)).

### DIFF
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -62,11 +62,7 @@ def login_user(user, remember=None):
         return False
 
     if _security.trackable:
-        if 'X-Forwarded-For' in request.headers:
-            remote_addr = request.headers.getlist(
-                "X-Forwarded-For")[0].rpartition(' ')[-1]
-        else:
-            remote_addr = request.remote_addr or 'untrackable'
+        remote_addr = request.remote_addr or 'untrackable'
 
         old_current_login, new_current_login = user.current_login_at, datetime.utcnow()
         old_current_ip, new_current_ip = user.current_login_ip, remote_addr
--- a/tests/test_trackable.py
+++ b/tests/test_trackable.py
@@ -9,11 +9,13 @@
 import pytest
 
 from utils import authenticate, logout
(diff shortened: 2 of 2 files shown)


lnielsen:
PyPy build is failing probably due to latest peewee release (2.8.2 release on Aug 9)

mattupstate:
This is great. Thanks for this.  Are you able to rebase off of develop and get the build passing?

jirikuncar:
Merged as 0676c29e4380c11f3ba0cff7b4af7958db8a11ab via #617.

PR #544, opened 2016-09-08
Include LICENSE in MANIFEST.in

pmlandwehr:
For the [conda-forge build](https://github.com/conda-forge/staged-recipes/pull/1515), we'd like to include a hard-link to the `LICENSE`. Doing this requires having an explicit reference to `LICENSE` in the manifest so it can be bundled with the source.


jonafato:
Thanks for the fix!

PR #548, opened 2016-09-20
Added support for lazy post logout view configuration

### LINKED ISSUE #499: Post login/logout inconsistent configuration
It's possible to set `SECURITY_POST_LOGIN_VIEW` before and after creating the `Security` class, however it's only possible to set `SECURITY_POST_LOGOUT_VIEW` before creating the security class.
Looking at the code, it seems the latter is being stored in an internal state, rather than queried from the application config.
I think the behavior should be consistent, and I'm leaning more towards the former behavior - meaning being able to set the configuration even after creating the `Security` class.

vToMy:
Fixes #499

### DIFF
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -246,6 +246,10 @@ def get_post_register_redirect(declared=None):
     return get_post_action_redirect('SECURITY_POST_REGISTER_VIEW', declared)
 
 
+def get_post_logout_redirect(declared=None):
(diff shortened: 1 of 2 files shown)


monkeysecurity:
More importantly, this fixes an open redirect bug in logout.

```
APP/logout?next=https://cnn.com/
```

Please merge this in ASAP and push out a new version.

jirikuncar:
@monkeysecurity thanks for the reminder. I have integrated it to our fork as https://github.com/inveniosoftware/flask-security-fork/pull/6.

Issue #552, opened 2016-10-16
FlaskWTFDeprecationWarning


swedishmike:
I get the following error when running a Flask app at the moment:

<- Cut ->
sphuc.py:34: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  form = NameForm()
<- Cut ->

Having looked at the issue it seems that the warning is coming from/due to flask-security, more specifically this line in forms.py:

`from flask_wtf import Form as BaseForm`

If I understand things correctly that should instead read:

`from flask_wtf import FlaskForm as BaseForm`

neaket360pi:
Looks like this was fixed in https://github.com/mattupstate/flask-security/pull/549

Can someone push up a new release to pypi?

jirikuncar:
@neaket360pi if you need a quick solution, you can use Flask-Security-Fork from PyPI (`pip uninstall Flask-Security`, `pip install Flask-Security-Fork` - no other changes needed).

neaket360pi:
Thanks @jirikuncar

mixmastamyk:
How bout it?  Tired of warning messages taking up the console.

jirikuncar:
closed via #553

PR #549, opened 2016-09-29
Bump Flask-WTF to 0.13

jacquerie:
We just began seeing all these warnings in our project:

```
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:89: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  class Form(BaseForm):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:154: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  class SendConfirmationForm(Form, UserEmailFormMixin):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:173: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  class ForgotPasswordForm(Form, UserEmailFormMixin):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:187: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  class PasswordlessLoginForm(Form, UserEmailFormMixin):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:204: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  class LoginForm(Form, NextFormMixin):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:251: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  UniqueEmailFormMixin, NewPasswordFormMixin):
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask_security/forms.py:256: FlaskWTFDeprecationWarning: "flask_wtf.Form" has been renamed to "FlaskForm" and will be removed in 1.0.
  NextFormMixin):
[...]

PR #553, opened 2016-10-16
Fix for issue #552 FlaskWTFDeprecationWarning

swedishmike:
This should solve the problem that I mentioned in issue  #552 FlaskWTFDeprecationWarning.

I just changed the import statement to match what the new 'demands' are.

### DIFF
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -12,7 +12,7 @@
 import inspect
 
 from flask import request, current_app, flash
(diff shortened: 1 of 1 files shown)


fuhrysteve:
See #549

neaket360pi:
This PR looks like a dup of https://github.com/mattupstate/flask-security/pull/549 and can be closed.

PR #565, opened 2016-11-15
security fix for changed email confirmation

slint:
When validating a confirmation token, also checks if the email that
the token was created with matches the user's current email.


jirikuncar:
It was released in Flask-Security-Fork v1.8.1 on November 15th, 2016.

PR #567, opened 2016-11-18
Add notes to documentation for login_user. Closes #566.

### LINKED ISSUE #566: TRACKABLE info is not stored when using login_user directly
We use
```
from flask_security import login_user
@app.route('/custom_login', methods=['POST'])
def custom_login():
    login_user(some_user)
    return redirect(...)
```
and noticed that information regarding current_login, last login, etc is not stored into the DB (sqlalchemy for us) unless we add a `db.session.commit()` after the `login_user(...)` call. 

After reading a bit the code, it looks like it could be fixed by adding a call to `after_this_request(db.session.commit)` in the `flask_securtity.utils.login_user` method. Not sure where we should add a unit test for this. Please if @mattupstate can hint us a bit on where is a good place to write the test for it, it would be great help to submit a PR and save you some work.

Thanks a lot.

avilaton:
When `SECURITY_TRACKABLE` is enabled, one must commit the transaction after the request in which `login_user` is called. This is used throughout `flask_security` and was learned from reading the code.

This also adds a test which shows how this works.

### DIFF
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -49,7 +49,8 @@ else:  # pragma: no cover
 
 
 def login_user(user, remember=None):
-    """Performs the login routine.
+    """Performs the login routine. If SECURITY_TRACKABLE is used, make sure you commit
+    changes after this request (i.e. ``app.security.datastore.commit()``).
 
     :param user: The user to login
     :param remember: Flag specifying if the remember cookie should be set. Defaults to ``False``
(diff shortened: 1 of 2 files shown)


avilaton:
@jirikuncar I tried at the time, it looks like those 7 are failing from before I changed anything. The helper `populate_data` from `tests/utils.py` blows up when running with pypy, not sure why.

jirikuncar:
It works well in the fork that includes various patches (see https://github.com/inveniosoftware/flask-security-fork/pull/4). I am going to merge. Thanks!

PR #575, opened 2017-01-04
Made email sending asynchronous

kantologist:


jirikuncar:
@kantologist if you need to send emails asynchronously, you can use `@security.send_mail_task` decorator and move the threading there.

--
see also https://pythonhosted.org/Flask-Security/customizing.html#emails-with-celery

PR #586, opened 2017-02-04
Fix login view filling

elysium31:
Fix login view to fill with request form data

### DIFF
--- a/flask_security/views.py
+++ b/flask_security/views.py
@@ -70,7 +70,7 @@ def login():
     if request.json:
         form = form_class(MultiDict(request.json))
     else:
-        form = form_class()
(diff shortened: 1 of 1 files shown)


flask_security/views.py
```diff
@@ -70,7 +70,7 @@ def login():
     if request.json:
         form = form_class(MultiDict(request.json))
     else:
-        form = form_class()
+        form = form_class(request.form)
```
jirikuncar: I am not really sure if we need it. Flask-WTF does it for us (https://flask-wtf.readthedocs.io/en/stable/api.html#flask_wtf.FlaskForm), but I can understand that it might be useful for custom form classes.

PR #589, opened 2017-02-08
Default SECURITY_EMAIL_SENDER respects MAIL_DEFAULT_SENDER in Flask-Mail

jxltom:
Since email is sent using Flask-Mail, when ```SECURITY_EMAIL_SENDER``` is not set for Flask-Security, it will be nice to search for ```MAIL_DEFAULT_SENDER``` setting, which sets default email sender in Flask-Mail, instead of using ```no-reply@localhost```.

In this case, users don't need to set same default sender as ```SECURITY_EMAIL_SENDER``` and ```MAIL_DEFAULT_SENDER``` for Flask-Security and Flask-Mail, respectively. On the other hand, users can explicitly set ```SECURITY_EMAIL_SENDER``` for Flask-Security and this setting will overide default sender in Flask-mail no matter it exists or not.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -425,10 +425,11 @@ class Security(object):
         """
         datastore = datastore or self.datastore
 
-        email_sender = app.config.get('MAIL_DEFAULT_SENDER')
-        _default_config['EMAIL_SENDER'] = email_sender or 'no-reply@localhost'
-
         for key, value in _default_config.items():
+            if key == 'EMAIL_SENDER' and app.config.get('MAIL_DEFAULT_SENDER'):
+                app.config.setdefault(
+                    'SECURITY_' + key, app.config.get('MAIL_DEFAULT_SENDER')
+                )
             app.config.setdefault('SECURITY_' + key, value)
 
         for key, value in _default_messages.items():
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -425,6 +425,9 @@ class Security(object):
         """
         datastore = datastore or self.datastore
 
+        email_sender = app.config.get('MAIL_DEFAULT_SENDER')
+        _default_config['EMAIL_SENDER'] = email_sender or 'no-reply@localhost'
+
         for key, value in _default_config.items():
             app.config.setdefault('SECURITY_' + key, value)
 


jirikuncar:
What about following magic? Maybe we can use `speaklater` instead of `LazyProxy`.

```diff
-     'EMAIL_SENDER': 'no-reply@localhost',
+     'EMAIL_SENDER': LazyProxy(lambda: current_app.config.get(
+         'MAIL_DEFAULT_SENDER', 'no-reply@localhost'
+     )),
```

jxltom:
@jirikuncar The magic is cool! The ```speaklater``` lib or ```LazyProxy``` from ```Flask-Babel``` (correct me if I am wrong) should works fine for that. But is that really necessary for adding a new library to fix this issue?

Actually ```werkzeug.local.LocalProxy``` can already do that and this won't introduce new library. Check [the code here](https://github.com/mattupstate/flask-security/compare/develop...jxltom:respect-flask-mail-localproxy).

But this has some side effects, e.g., following code has to be run under application context otherwise error will raised.

```python
from flask_security.utils import get_config
get_config(app)['EMAIL_SENDER']
```

So I am not sure which way is better. For now, I still keep this PR to be merged since it will not change other functions behavior and introducing new libs.

flask_security/core.py
```diff
@@ -425,6 +425,9 @@ def init_app(self, app, datastore=None, register_blueprint=True,
         """
         datastore = datastore or self.datastore
 
+        email_sender = app.config.get('MAIL_DEFAULT_SENDER')
+        _default_config['EMAIL_SENDER'] = email_sender or 'no-reply@localhost'
```
jirikuncar: We should not modify the `_default_config`, but instead directly `setdefault` on `app.config` with new value.

```python
# add after _default_config
_default_config_from_app = {
    'MAIL_DEFAULT_SENDER': 'EMAIL_SENDER',
}

for app_key, key in _default_config_from_app.items():
    value = app.config.get(app_key)
    if value:
        app.config.setdefault('SECURITY_' + key, value)
```

jxltom: Yes, you are right. Modifying ```_default_config``` may initialize other apps incorrectly. I have updated the PR.

PR #591, opened 2017-02-10
Add validators to default LoginForm

jxltom:
Data required validators are added to several fields since these fields are always essencial. Then invalid requests with empty fields won't hit server side and website users should have better experiences.

- Data required valildators are added to ```email``` and ```password``` fields of ```LoginForm``` since they are always needed for login.

-  Data required valildators are added to ```new_password_confirm``` and ```password_confirm``` in ```ChangePasswordForm``` and ```PasswordConfirmFormMixin``` respectively since reconfirmed password are also always neccessary.

### DIFF
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -205,8 +205,9 @@ class PasswordlessLoginForm(Form, UserEmailFormMixin):
 class LoginForm(Form, NextFormMixin):
     """The default login form"""
 
-    email = StringField(get_form_field_label('email'))
-    password = PasswordField(get_form_field_label('password'))
+    email = StringField(get_form_field_label('email'), validators=[Required()])
+    password = PasswordField(get_form_field_label('password'),
+                             validators=[password_required])
     remember = BooleanField(get_form_field_label('remember_me'))
     submit = SubmitField(get_form_field_label('login'))
 
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -127,7 +127,8 @@ class NewPasswordFormMixin():
 class PasswordConfirmFormMixin():
     password_confirm = PasswordField(
         get_form_field_label('retype_password'),
-        validators=[EqualTo('password', message='RETYPE_PASSWORD_MISMATCH')])
+        validators=[password_required,
+                    EqualTo('password', message='RETYPE_PASSWORD_MISMATCH')])
 
 
 class NextFormMixin():
@@ -285,8 +286,8 @@ class ChangePasswordForm(Form, PasswordFormMixin):
 
     new_password_confirm = PasswordField(
         get_form_field_label('retype_password'),
-        validators=[EqualTo('new_password',
-                            message='RETYPE_PASSWORD_MISMATCH')])
+        validators=[password_required,
+                    EqualTo('new_password', message='RETYPE_PASSWORD_MISMATCH')])
 
     submit = SubmitField(get_form_field_label('change_password'))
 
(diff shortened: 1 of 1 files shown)


flask_security/forms.py
```diff
@@ -203,8 +204,9 @@ def validate(self):
 class LoginForm(Form, NextFormMixin):
     """The default login form"""
 
-    email = StringField(get_form_field_label('email'))
```
jirikuncar: Can you have a look at code on line 223 if we still need it when using validators?

```python
     if self.email.data.strip() == '':
             self.email.errors.append(get_message('EMAIL_NOT_PROVIDED')[0])
             return False
 
         if self.password.data.strip() == '':
             self.password.errors.append(get_message('PASSWORD_NOT_PROVIDED')[0])
             return False
```

flask_security/forms.py
```diff
@@ -125,7 +125,8 @@ class NewPasswordFormMixin():
 class PasswordConfirmFormMixin():
     password_confirm = PasswordField(
         get_form_field_label('retype_password'),
-        validators=[EqualTo('password', message='RETYPE_PASSWORD_MISMATCH')])
+        validators=[password_required,
```
jirikuncar: IMHO `password_required` is not needed since the password must be equal to `password` field which is already required.

flask_security/forms.py
```diff
@@ -274,7 +276,8 @@ class ChangePasswordForm(Form, PasswordFormMixin):
 
     new_password_confirm = PasswordField(
         get_form_field_label('retype_password'),
-        validators=[EqualTo('new_password', message='RETYPE_PASSWORD_MISMATCH')])
+        validators=[password_required,
```
jirikuncar: similar to https://github.com/mattupstate/flask-security/pull/591/files#r100505404

flask_security/forms.py
```diff
@@ -203,8 +204,9 @@ def validate(self):
 class LoginForm(Form, NextFormMixin):
```
jirikuncar: I am not sure why we can't use `EmailFormMixin` and `PasswordFormMixin` here too.

flask_security/forms.py
jxltom: I don't thinks we need it. But it looks like deleting it will break some tests since they will test the responses are matched to these error strings when invalid email or password form data are posted.

flask_security/forms.py
jxltom: But one may give the password yet forget to reconfirm it. In this case, a request can still be posted to the server. By providing a data required attributes, the browser should remind the users that the reconfirm password field is empty. This provides better user experience and save server resources by filtering invalid posts before sending.

flask_security/forms.py
jxltom: It looks like we can but I‘am not sure about that.

flask_security/forms.py
jirikuncar: You might need to include error message for the new validators.

biomap:
Does this take into account if you use an alternative identifier instead or in addition to email??  I have an app right now that uses username only for auth and these changes would break it.

jirikuncar:
@biomap can you elaborate more on it or provide a little snippet? IMHO it's only changing the form validation so the fields are correctly marked as required.

biomap:
I will have to look at it closer but it appears as if you are going to require a field called email which will then run through the email validator hence mandating if conforms with all email validators.  ie a login without an @ would be rejected if using just a username.  Again I will look at it closer but we should be extra careful adding required or mandated fields.

jirikuncar:
@biomap you can always override the form, but maybe we can think about calling the field "identifier" or "id" as generic name for `find_user` method.

flask_security/forms.py
jirikuncar: @jxltom thank you for the clarification. Can you keep the `EqualTo` validator first so user get expected message  `RETYPE_PASSWORD_MISMATCH`?

PR #592, opened 2017-02-10
Set anonymous required for send_confirmation

jxltom:
If a user has already signed in, it means the email has already been confirmed if confirmable is true and there is no point in allowing website users to access resending confimation page anymore. So this ```send_confirmation``` function which provides ```/confirm``` view should be anonymous required.

If this setting is not set by design for some purpuses, I will keep this PR here.

### DIFF
--- a/flask_security/views.py
+++ b/flask_security/views.py
@@ -183,6 +183,7 @@ def token_login(token):
     return redirect(get_post_login_redirect())
 
 
+@anonymous_user_required
 def send_confirmation():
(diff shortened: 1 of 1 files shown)


jirikuncar:
IMHO one can request confirmation email from profile page e.g. after changing email address. Hence we should allow sending confirmation email even for signed in users.

jxltom:
@jirikuncar Aha, that makes sense. I closed it.

PR #593, opened 2017-02-10
Update docs on setting send_mail_task using factory method

jxltom:
#194 

@jirikuncar I did't attach the ```flask_mail.Message``` can not be jsonified part since that looks like more related to ```Celery```. Let me know if there is anything inappropriate.

### DIFF
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -160,4 +160,41 @@ decorator like so::
     def delay_security_email(msg):
         send_security_email.delay(msg)
 
+If factory method is going to be used for initialization, use ``_SecurityState``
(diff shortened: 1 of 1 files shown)


jirikuncar:
@jxltom it would be interesting to include the custom serialization of `Message` as presented before:

```python
@celery.task
def send_flask_mail(**kwargs):
        mail.send(Message(**kwargs))

@security_ctx.send_mail_task
def delay_flask_security_mail(msg):
    send_flask_mail.delay(subject=msg.subject, sender=msg.sender,
                          recipients=msg.recipients, body=msg.body,
                          html=msg.html)
```

jxltom:
@jirikuncar Sure, already done. xD

PR #594, opened 2017-02-10
Custom User Payload Handler

marcosschroh:
Sometimes after Create or Login an user (base on Token) we need some information about he/she, so after receive the token we need to request again the server.

Example: 
When we send the email and password Flask-Security returns:

```python
{
    "meta": {
        "code": 200
    },
    "response": {
        "user": {
            "authentication_token": "WyI5OTk5OTk5IiwiMTE0ODZlNzU3MGNjNWRkMjFiZDhkMzI4M2I5ZTUxZjYiXQ.C39mEg.UGm6QqOaGlZ3wJsu5jQuWaIIA8M",
            "id": "9999999"
        }
    }
}
```

Is really good. But if we need more information we should request the server again.

I introduced a new setting flag called SECURITY_CUSTOM_USER_PAYLOAD. By default is False.
If you set to True, you must define a function in your User Model called 'payload_handler' that must return a serializable object:

```python
UserModel(Base, UserMixin):
    id = db.Column(db.Integer, primary_key=True)
    email = db.Column(db.String(120), unique=True, nullable=False)
    password = db.Column(db.String(255))
    name = db.Column(db.String(80))
    ....
    
    def payload_handler(self):
        return {
            id: self.id,
            name: self.name,
            email: self.email  
        }
```

Now after login the user the response is:

```python
{
    "meta": {
        "code": 200
    },
    "response": {
        "user": {
            "authentication_token": "WyI5OTk5OTk5IiwiMTE0ODZlNzU3MGNjNWRkMjFiZDhkMzI4M2I5ZTUxZjYiXQ.C39maw.EYNyKu2-T70BJFyWlFiMsjWh-XA",
            "email": "test@tes.com",
            "id": 9999999,
            "name": "test",
        }
    }
}
```

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -57,7 +57,7 @@ If you want a custom payload after Register or Login an user, define
 the method get_security_payload in your User model. The method must return a
 serializable object:
 
-::
+.. code-block:: python
 
     class User(db.Model, UserMixin):
         id = db.Column(db.Integer, primary_key=True)
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -58,16 +58,19 @@ the method get_security_payload in your User model. The method must return a
 serializable object:
 
 ::
-    UserModel(Base, UserMixin):
+
+    class User(db.Model, UserMixin):
         id = db.Column(db.Integer, primary_key=True)
-        email = db.Column(db.String(120), unique=True, nullable=False)
-        password = db.Column(db.String(255))
+        email = TextField()
+        password = TextField()
+        active = BooleanField(default=True)
+        confirmed_at = DateTimeField(null=True)
         name = db.Column(db.String(80))
-        ....
-        
+
+        # Custom User Payload
         def get_security_payload(self):
             return {
-                id: self.id,
-                name: self.name,
-                email: self.email  
+                'id': self.id,
+                'name': self.name,
+                'email': self.email
             }
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -49,3 +49,25 @@ additional fields:
 * ``last_login_ip``
 * ``current_login_ip``
 * ``login_count``
+
+Custom User Payload
+^^^^^^^^^^^^^^^^^^^
+
+If you want a custom payload after Register or Login an user, define
+the method get_security_payload in your User model. The method must return a
+serializable object:
+
+::
+    UserModel(Base, UserMixin):
+        id = db.Column(db.Integer, primary_key=True)
+        email = db.Column(db.String(120), unique=True, nullable=False)
+        password = db.Column(db.String(255))
+        name = db.Column(db.String(80))
+        ....
+        
+        def get_security_payload(self):
+            return {
+                id: self.id,
+                name: self.name,
+                email: self.email  
+            }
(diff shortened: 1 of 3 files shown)


flask_security/views.py
```diff
@@ -45,6 +45,8 @@ def _render_json(form, include_user=True, include_auth_token=False):
         response = dict()
         if include_user:
             response['user'] = dict(id=str(form.user.id))
```
jirikuncar: What about moving this code to default implementation in `flask_security.code:UserMixin.get_security_payload()`.

```python
class UserMixin(BaseUserMixin):
    # ...
    def get_security_payload(self):
        """Serialize user object as response payload."""
        return {'id': str(self.id)}
```

```diff
-             response['user'] = dict(id=str(form.user.id))
+            response['user'] = form.user.get_security_payload()
```

Please make sure that you include tests and documentation for this new feature.

marcosschroh: Yes, you are right. I think is a better solution because we don't have the necessity to define another setting in the config. If you want a different payload you only need to rewrite the method get_security_payload() in the User Model. 

So, in the view.py file we get the payload:

```python
response = dict()
if include_user:
    response['user'] = form.user.get_security_payload()
```

```python
class UserMixin(BaseUserMixin):
    # ...
    def get_security_payload(self):
        """Serialize user object as response payload."""
        return {'id': str(self.id)}
```

If you need a different Payload redifine the method: 

```python
UserModel(Base):
    id = db.Column(db.Integer, primary_key=True)
    email = db.Column(db.String(120), unique=True, nullable=False)
    password = db.Column(db.String(255))
    name = db.Column(db.String(80))
    
    def get_security_payload(self):
        return {
            id: self.id,
            name: self.name,
            email: self.email  
        }
```

jirikuncar: @marcosschroh can you include your last comment in docs? I think it will be very helpful. Thanks!

docs/models.rst
```diff
@@ -49,3 +49,25 @@ additional fields:
...
+
+Custom User Payload
+^^^^^^^^^^^^^^^^^^^
+
+If you want a custom payload after Register or Login an user, define
+the method get_security_payload in your User model. The method must return a
+serializable object:
+
+::
+    UserModel(Base, UserMixin):
```
jirikuncar:
```python
class UserModel(Base, UserMixin):
```

docs/models.rst
```diff
@@ -49,3 +49,25 @@ additional fields:
...
 * ``login_count``
+
+Custom User Payload
+^^^^^^^^^^^^^^^^^^^
+
+If you want a custom payload after Register or Login an user, define
+the method get_security_payload in your User model. The method must return a
+serializable object:
+
+::
```
jirikuncar:
```
.. code-block: python
```

docs/models.rst
```diff
@@ -49,3 +49,25 @@ additional fields:
...
+    UserModel(Base, UserMixin):
+        id = db.Column(db.Integer, primary_key=True)
+        email = db.Column(db.String(120), unique=True, nullable=False)
+        password = db.Column(db.String(255))
+        name = db.Column(db.String(80))
+        ....
```
jirikuncar: IMHO we can remove the dots so one can simply copy & paste the example.

marcosschroh:
@jirikuncar I removed the dots in the previous commit. Is there anything else that I should need to change?

docs/models.rst
jirikuncar: ping @marcosschroh

Issue #604, opened 2017-03-13
login with numeric field


stamaimer:
I have read #124 #396 #464. And I use the method mentioned above to login user with phone number. After read the `get_user` method of `SQLAlchemyUserDatastore`. I found `get_user` try to use `_is_numeric` transform `identifier`(the value of email key in login form) to integer. If the identifier can transform to a integer, it will use as the primary key to get user from user table. I don't know why do this before `get_identity_attributes`. I found this will cause bug if someone want to use phone number as the user identity.

biomap:
A phone number would still be a string as it has numeric and has punctuation '555-555-5555' or something as such.  if you are trying to login with JUST a numeric value then over class (super) the UserDatastore and replace the get_user method.  Otherwise this is a a specific call to get the ID (Primary Key) of the database row and does a ...query.get(ID) call to the database (in SQLA at least).  This is a fundamental call to retrieve a one specific the database object.  Your case is edge and prob best served by using an alternative UserDatastore or super classing the in-built datastore and replacing just that one method to suit your needs.

good luck :)

stamaimer:
@biomap Thanks for your reply. In China there is only digits in phone number. And i can't find where use the `get_user` by primary key in the project. And i also find `find_user` in `UserStore` or subclasses.

devrandom:
```
    def _is_numeric(self, value):
        try:
            int(value)
        except (TypeError, ValueError):
            return False
        return True
```

This is bad from several points of view.  It means that emails with leading digits will fail to log in.  It may potentially lead to a security issue, if the app retrieves a different user than expected.

devrandom:
`isinstance(value, int)` would be the right thing to do here

devrandom:
alternatively, if you want to support strings, then match against `/^\d+$/`

jirikuncar:
@devrandom can you please submit a PR? You can start from #633. Thanks!

PR #608, opened 2017-03-22
prevent convert a number string of identity attribute to number and p…

stamaimer:
Prevent convert a number string of identity attribute to number and prevent get user by primary id. For example, use phone of user as the identity attribute. #604


jirikuncar:
@stamaimer please check #633. You can of course create your own datastore that will handle this specific problem until we find widely acceptable solution.

Issue #559, opened 2016-10-28
Integration and release processes.


jirikuncar:
Dear @mattupstate and @jonafato:

In the light of recent Flask-Login release, I would like to ask you about plans for this package. There are many useful PRs that deserve to be merged (#494 #502 #519 #526 #528 #537 #538 #542 #544 #549 and more). Some of them are over a year old without any comment from maintainers.

If there is an interest I would offer my help with maintaining this package. Until the reply I am going to manage our [fork](https://github.com/inveniosoftware/flask-security-fork) so we can take benefits of the latest comunity work in both Flask-Login and Flask-Security packages.

Kind regards,
Jiri

fuhrysteve:
It would be great if @jirikuncar could be added as a maintainer to help catch up. I don't see any value in having two forks, especially for such a small (albeit active) community.

melonista:
IMO, if current maintainers are not interested on this, @jirikuncar (or whoever) should convert the fork into an independent library. This library is very useful and should be up to date.

jonafato:
We're working on a maintenance path forward. I'll post another update here with details this week.

chris-allan:
It would be great if the maintainership could be further clarified so that those of us who depend on this great library can decide what to do. We've been using @jirikuncar's fork for almost 6 weeks now. Especially as it includes #565.

All of us have families, day jobs and other distractions in our lives but it has now been over a month. As @jirikuncar has already stated, there are numerous useful PRs that either deserve to be merged or at least commented upon. Contributor momentum is being lost and it would be a real shame if people were unknowingly exposed to security vulnerabilities because they weren't using the fork.

jirikuncar:
Dear all,
I have tried to integrate as many as possible bugfixes in past months. It is time to move forward and bring new features too (e.g. CLI using Click https://github.com/inveniosoftware/flask-security-fork/pull/9). Since there is no update from current maintainers, I will continue development in our [fork](https://github.com/inveniosoftware/flask-security-fork) hoping that one day it can become the official "master" branch here.

Wishing all the best in 2017!

mattupstate:
Just want to mention that @jonafato and I will be meeting next week to work out a maintenance plan.

fuhrysteve:
I hope you guys are able to come up with a way to keep the community united!

jirikuncar:
@mattupstate @jonafato do you have any updates on a maintenance plan?

mattupstate:
[...]

PR #610, opened 2017-03-25
Replace md5 hashing with sha512

ioparaskev:
Use sha512 instead of md5 for any hashing operation. MD5 is completely broken and should not be used for anything related to password. Even using MD5 to share password hashes should be treated as if giving the hashes in plaintext.

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -18,9 +18,7 @@ from itsdangerous import URLSafeTimedSerializer
 from passlib.context import CryptContext
 from werkzeug.datastructures import ImmutableList
 from werkzeug.local import LocalProxy
-from werkzeug.security import safe_str_cmp
-
-from .utils import config_value as cv, get_config, sha512, url_for_security, string_types
+from .utils import config_value as cv, get_config, hash_data, verify_hash, url_for_security, string_types
 from .views import create_blueprint
 from .forms import LoginForm, ConfirmRegisterForm, RegisterForm, \
     ForgotPasswordForm, ChangePasswordForm, ResetPasswordForm, \
@@ -198,7 +196,7 @@ def _token_loader(token):
         data = _security.remember_token_serializer.loads(
             token, max_age=_security.token_max_age)
         user = _security.datastore.find_user(id=data[0])
-        if user and safe_str_cmp(sha512(user.password), data[1]):
+        if user and (verify_hash(data[1], user.password)):
             return user
     except:
         pass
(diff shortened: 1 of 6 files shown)


jirikuncar:
@ioparaskev I agree we should not use `md5` for password related operations. In case of tokens, the `md5` is used to hash password hashes so it should not be an immediate thread. Can you update the PR in a way in which it can handle tokens generated with both hashing functions similarly to deprecated password hashing functions (see https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L249-L257)?

Maybe something along these lines:
```python
from passlib.context import CryptContext
ctx = CryptContext(schemes=['hex_sha256', 'hex_md5'], deprecated=['hex_md5'])
ctx.verify(user.password, password_hash)
```

ioparaskev:
@jirikuncar It's not an immediate risk, although using md5 for operations where a password hash is involved is like sending the password hash. MD5 hashing a password hash does not add any security at all. So since we don't want the password hashes to be publicly available, we should also not send them hashed with md5.

I was also thinkg instead of the hexdigest, to use the salted version to avoid precomputed attacks.
So I was thinking something like the following:
````python
from passlib.context import CryptContext
ctx = CryptContext(schemes=['sha512_crypt', 'hex_md5'], deprecated=['hex_md5'])
#etc...
````

**Comment by jirikuncar** [comment]
> I was also thinkg instead of the hexdigest, to use the salted version to avoid precomputed attacks.

@ioparaskev Sounds good 👍

File: flask_security/core.py [hunk 1]
```diff
@@ -198,7 +196,7 @@ def _token_loader(token):
         data = _security.remember_token_serializer.loads(
             token, max_age=_security.token_max_age)
         user = _security.datastore.find_user(id=data[0])
-        if user and safe_str_cmp(md5(user.password), data[1]):
+        if user and (verify_hash(data[1], user.password)):
```
jirikuncar: You can remove the bracket around the `verify_hash` call.

flask_security/utils.py
```diff
@@ -163,8 +164,17 @@ def encode_string(string):
     return string
 
 
-def md5(data):
-    return hashlib.md5(encode_string(data)).hexdigest()
+def hash_data(data):
+    ctx = CryptContext(schemes=['sha512_crypt', 'hex_md5'],
+                       deprecated=['hex_md5'])
+    return ctx.hash(encode_string(data))
+    # return hashlib.sha512(encode_string(data)).hexdigest()
```
jirikuncar: Please remove the commented line. Thank you.

ioparaskev:
do you think anything else is needed?

ioparaskev:
@jirikuncar do you think anything else is needed

jirikuncar:
@ioparaskev it has been integrated to Flask-Security-Fork as https://github.com/inveniosoftware/flask-security-fork/pull/30. There was no answer from current maintainers in https://github.com/mattupstate/flask-security/issues/559 so I can't merge it here 😞

PR #614, opened 2017-04-16
 Make explicit support for both Python 2 and 3

cclauss:


cclauss:
Switching over to https://pypi.python.org/pypi/Flask-Security-Fork

Issue #601, opened 2017-03-07
SECURITY_CONFIRMABLE mixup between confirm password and confirm email


sdiepend:
I think there is some weird behaviour when using SECURITY_CONFIRMABLE.

The documentation states the following, so this is a flag to enable confirmation of an email address.

> SECURITY_CONFIRMABLE 	
> Specifies if users are required to confirm their email address when registering a new account. If this value is True, Flask-Security creates an endpoint to handle confirmations and requests to resend confirmation instructions. The URL for this endpoint is specified by the SECURITY_CONFIRM_URL configuration option. Defaults to False.

But when set to True it also removes the second password field, so I think this is mixing up email confirimation and password confirmation. 

register_user.html
```
{% if register_user_form.password_confirm %}
    {{ render_field_with_errors(register_user_form.password_confirm) }}
  {% endif %}
```

I think it has to do with the following two forms:
forms.py
```
class ConfirmRegisterForm(Form, RegisterFormMixin,
                          UniqueEmailFormMixin, NewPasswordFormMixin):
    pass


class RegisterForm(ConfirmRegisterForm, PasswordConfirmFormMixin,
                   NextFormMixin):
    def __init__(self, *args, **kwargs):
        super(RegisterForm, self).__init__(*args, **kwargs)
        if not self.next.data:
            self.next.data = request.args.get('next', '')
```

views.py
```
def register():
    """View function which handles a registration request."""

    if _security.confirmable or request.json:
        form_class = _security.confirm_register_form
    else:
        form_class = _security.register_form

```

biomap:
Two completely different things register_user_form.password_confirm is a form validator.
https://github.com/mattupstate/flask-security/blob/develop/flask_security/forms.py#L125
security.confirmable (SECURITY_CONFIRMABLE=True) allows you to send confirmation emails before the user is activated.

sdiepend:
Hello, why wontfix? I can have a go at it this weekend/next weekend if you want.

sdiepend:
@jirikuncar I almost have a fix ready. But it doesn't give you the option to turn off retyping the password. (password confirmation).

jirikuncar:
@sdiepend please check #580 since it looks very similar.

sdiepend:
@jirikuncar I would add a SECURITY_RETYPABLE to add a flag to enable/disable the retyping of the password. But I'm not familiar with the testing setup, would you maybe have some time some day to walk me through?

jirikuncar:
[...]

Issue #580, opened 2017-01-10
Customizing register view switches which form class is used unexpectedly


UnicycleMatt:
1. I've created my own ExtendedRegisterForm class.

class ExtendedRegisterForm(RegisterForm):
	first_name = StringField('First Name', [validators.required()])
	last_name = StringField('Last Name', [validators.required()])
	phone_number = StringField('Phone Number', [validators.required()])

2. I am passing this into the Security object as such:

security = Security(app, user_datastore, register_form=ExtendedRegisterForm)

3. Concurrently, I have SECURITY_CONFIRMABLE set to True as I want to do email confirmation AND avoid having the password confirm field. I want to minimize unnecessary UI fields as you mention with this flag.

SECURITY_CONFIRMABLE = True

4. Finally, I've added new template variables to show my form inputs.

{{ render_field_with_errors(register_user_form.first_name, class_="form-control input-lg", placeholder="First Name") }}
{{ render_field_with_errors(register_user_form.last_name, class_="form-control input-lg", placeholder="Last Name") }}
{{ render_field_with_errors(register_user_form.phone_number, class_="form-control input-lg", placeholder="Phone Number") }}

5. However, now when I render my template, I get an exception stating that the ConfirmRegisterForm does not have any of these fields. It appears that Flask-Security is now choosing to render the ConfirmRegisterForm instead of the RegisterForm override. If I change my Security initializer, the form renders, but of course has the extra confirm password field.

security = Security(app, user_datastore, register_form=ExtendedRegisterForm, confirm_register_form=ExtendedRegisterForm)

The expected behavior would be to show the ExtendedRegisterForm in place of RegisterForm instead of ConfirmRegisterForm, and for the template to avoid having that password confirm field.

biomap:
#601

PR #637, opened 2017-05-13
Fixing the different possibilities for the register form.

sdiepend:
Fix for issues
https://github.com/mattupstate/flask-security/issues/601
https://github.com/mattupstate/flask-security/issues/580

Adds a new flag: SECURITY_RETYPABLE which allows you to specify
if users have to confirm(retype) their password or not when registering.

Now you have four different combinations of possible register forms.

CONFIRMABLE: False, RETYPABLE: False => RegisterForm
CONFIRMABLE: True, RETYPABLE: False  => ConfirmRegisterForm
CONFIRMABLE: False, RETYPABLE: True  => RetypeRegisterForm
CONFIRMABLE: True, RETYPABLE: True   => ConfirmRetypeRegisterForm

Only when using the API it will default to ConfirmRegisterForm as before.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -180,6 +180,8 @@ Feature Flags
                           instructions. The URL for this endpoint is specified
                           by the ``SECURITY_CONFIRM_URL`` configuration option.
                           Defaults to ``False``.
+``SECURITY_RETYPABLE``    Specifies if users have to retype/confirm their 
+                          password when registering. Defaults to ``False``.
 ``SECURITY_REGISTERABLE`` Specifies if Flask-Security should create a user
                           registration endpoint. The URL for this endpoint is
                           specified by the ``SECURITY_REGISTER_URL``
--- a/docs/customizing.rst
+++ b/docs/customizing.rst
@@ -68,7 +68,9 @@ Forms
 
 All forms can be overridden. For each form used, you can specify a
 replacement class. This allows you to add extra fields to the
-register form or override validators::
+register form or override validators, make sure to inherit from the
+correct class when you have set SECURITY_CONFIRMABLE and/or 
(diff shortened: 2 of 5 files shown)


jirikuncar:
This functionality can be done already via configurable forms (see https://github.com/mattupstate/flask-security/issues/601#issuecomment-299401601). Also I see the different forms for regular and API confusing.

I would propose that we can split the forms so one can easily choose one with password confirmation or without it. WDYT?

sdiepend:
Yes, split them, but we should add a flag for this. That's the SECURITY_RETYPABLE I had chosen. Because SECURITY_CONFIRMABLE is used for account/email confirmation.

We probably don't need a different form for the account/email confirmation as has no influence on the fields of the form itself.

jirikuncar:
The functionality added by `SECURITY_RETYPABLE` flag can be configured only via `SECURITY_ REGISTER_FORM` and `SECURITY_CONFIRM_REGISTER_FORM`, hence I am not a big fan of adding it.

sdiepend:
@jirikuncar I can't find the flags/config you are describing 
https://github.com/mattupstate/flask-security/search?utf8=%E2%9C%93&q=SECURITY_REGISTER_FORM&type=
https://github.com/mattupstate/flask-security/search?q=SECURITY_CONFIRM_REGISTER_FORM&type=Code&utf8=%E2%9C%93

jirikuncar:
Check:
- https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L343-L345
- https://github.com/mattupstate/flask-security/blob/develop/tests/test_misc.py#L256-L276

We need to document it better O:-)

jirikuncar:
@sdiepend feel free to re-open the PR if you are not happy with the proposed solution.

PR #642, opened 2017-05-18
Do not use strip() to compare passwords in flask_security.forms.ChangePasswordForm

Devking:
**Problem**
In the password change form (`flask_security.forms.ChangePasswordForm`), `strip()` is used to compare old and new passwords. However, whitespace is an accepted part of a password (as long as it's accompanied by other characters). For example, '     a' (six leading spaces) and 'a     ' (six trailing spaces) are two different and both *valid* passwords. Indeed, Flask-Security allows for both as valid passwords to be used, and they are not interchangeable on login. However, because the `strip()` is currently being done, on the password change form, you incorrectly get the message that the two distinct passwords are the same.

**Solution**
`strip()` is no longer used to compare if an old password and a new password are the same.

**Miscellanea**
As for the philosophical argument of whether you *should* do whitespace stripping on passwords, I'm of the camp that you *should not*. Similar sentiments can be found [here](http://stackoverflow.com/a/7240898) and [here](https://ux.stackexchange.com/a/8011), though these are not at all authoritative sources. 

However, the above argument doesn't matter as much here, since Flask-Security does take its own stance already &ndash; not to strip whitespace on passwords. The stripping does not occur on registration, nor on login. To follow suit, it should not be done in the password change form.

### DIFF
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -291,7 +291,7 @@ class ChangePasswordForm(Form, PasswordFormMixin):
         if not verify_and_update_password(self.password.data, current_user):
             self.password.errors.append(get_message('INVALID_PASSWORD')[0])
             return False
-        if self.password.data.strip() == self.new_password.data.strip():
+        if self.password.data == self.new_password.data:
             self.password.errors.append(get_message('PASSWORD_IS_THE_SAME')[0])
             return False
         return True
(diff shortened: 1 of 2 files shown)


Devking:
And apologies if this PR does not fit the conventions being followed by this repo; I couldn't find a `CONTRIBUTING.md` guide (or equivalent). I'm happy to make changes to fit existing conventions, if needed.

jirikuncar:
@Devking can you please include a test case with new password containing a whitespace?

Devking:
@jirikuncar good looks on the test case request. I've added one test case to cover the change, which would previously fail but with the change will pass. Ready for re-review.

PR #651, opened 2017-06-04
Make user inactive on register if _security.confirmable

Jaza:
Without this fix, current behaviour is: user registers; user is immediately set to `active=True` and `confirmed_at=now` in DB; user clicks link in confirmation email; user sees "email already confirmed" warning message; nothing changes for user's record in DB.

With this fix, behaviour is as expected: user registers; user is set to `active=False` and `confirmed_at=None` in DB; user clicks link in confirmation email; user sees "thanks for confirming your email" success message; user's record in DB is updated to have `active=True` and `confirmed_at=now`.

### DIFF
--- a/flask_security/views.py
+++ b/flask_security/views.py
@@ -115,7 +115,11 @@ def register():
     form = form_class(form_data)
 
     if form.validate_on_submit():
-        user = register_user(**form.to_dict())
+        create_user_kwargs = form.to_dict()
+        if _security.confirmable:
+            create_user_kwargs['active'] = False
+
+        user = register_user(**create_user_kwargs)
         form.user = user
 
         if not _security.confirmable or _security.login_without_confirmation:


jonafato:
`active` is a separate concept from confirmable here and comes from [Flask Login](https://flask-login.readthedocs.io/en/latest/#your-user-class); it's meant to be used for things like disabling or suspending accounts. The Flask Security login form should currently [prevent](https://github.com/mattupstate/flask-security/blob/fc8d27fe41cdc2d1028e1e89d6315e8ba9396e77/flask_security/forms.py#L244-L246) any non-confirmed users from logging in.

What does your UserDatastore model look like here? Are you setting a default for `confirmed_at` by any chance?

Jaza:
@jonafato: yes, I have tied `active` and `confirmed_at` together by way of this code:

    def update_confirmedat_before_save(mapper, connection, target):
        """Event listener callback to update the confirmed at field."""
        if target.active and not target.confirmed_at:
            target.confirmed_at = datetime.utcnow()
        elif target.confirmed_at and not target.active:
            target.active = True

    event.listen(User, 'before_insert', update_confirmedat_before_save)
    event.listen(User, 'before_update', update_confirmedat_before_save)

I didn't realise that Flask-Security expects `active` to specifically only indicate whether a user account is blocked, I thought it could also indicate whether a user account is confirmed, without other side effects.

So, with the code that I currently have in place, `confirmed_at` gets set on user register because Flask-Security is setting `active` to True. Looks like I'll have to reconsider this code on my end.

jonafato:
Flask Security sets `confirmed_at` [here](https://github.com/mattupstate/flask-security/blob/fc8d27fe41cdc2d1028e1e89d6315e8ba9396e77/flask_security/views.py#L233-L237) inside the `confirm_email` view. The intent of this workflow is to ensure that the registered user has access to the email they're signing up with. The process goes like:

1. User signs up with email address.
2. Flask Security sends a unique confirmation link to that email address.
3. User follows the confirmation link, and Flask Security sets the `confirmed_at` datetime attribute on the user object.

When the confirmable option is enabled, Flask Security's login function prevents the user from logging in before they've completed step 3. If `confirmed_at` is set outside of this process, there's no guarantee that the user was actually able to receive email at their address, so the feature is effectively circumvented.

If your goal is to prevent users who have not confirmed their email addresses from logging in, the standard behavior should work for you without modification. Removing your event handler should get your application to that state.

Jaza:
Yes, removing the event handler fixes it for me.

This happened for me in the first place, because I have added `active` and `confirmed_at` to various other SQLAlchemy models (other than `User`) in my app. In those other models, I treat `active` as `is_published`, and `confirmed_at` as `published_at`. For those models, my event handler makes sense. My mistake was adding the same event handler to `User`, and not checking how Flask-Security manages the value of `active` and `confirmed_at` during the registration / confirmation process.

PR #653, opened 2017-06-06
Release v3.0.0

### LINKED ISSUE #644: Update pypi to version 2
Current version on [pypi](https://pypi.python.org/pypi/Flask-Security) is 1.7.5 and miss new functions such as i18n, etc..

It would be nice to be able to install releases from pypi or at least have git tags to use with pip.

jirikuncar:
(closes #644)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: python
 
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
   - "3.6"
   - "pypy"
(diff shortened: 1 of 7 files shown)


avilaton:
I see that 3.0.0 is released but documentation at https://flask-security.readthedocs.io is still old, any change we get a new docs build?

jirikuncar:
@avilaton I have added `3.0.0` and `develop` builds on readthedocs.

avilaton:
Thanks @jirikuncar, I appreciate that. It is still not clear to me why 3.0.0 isn't the actual latest release.
We were looking at the code to see if it should be the case that a `user.is_active=False` value should actually log the user out or not.

It appears to be that nothing is done with a user account which is deactivated after logging in. We want to actually kick the user out on the next request if the account is deactivated while the user is logged in.

We were reading the code and found https://github.com/mattupstate/flask-security/blob/develop/docs/customizing.rst#authorization-with-oauth2
and found the override to _get_login_manager could be a nice place for that, but it is not that clean cut in 1.7.5 nor in 3.0.0.

Any chance we get what is currently in `develop` to be released as 3.0.1?

jirikuncar:
> Any chance we get what is currently in develop to be released as 3.0.1?

@avilaton I will need to find some time to compile release notes. Any help is appreciated ;) There were some bigger changes and config options which might deserve to bump version to `3.1.0`.

avilaton:
Took a shot at drafting CHANGES from 3.0.0 to current develop, not sure how to prune that list but it should be a complete list of changes. 
https://github.com/mattupstate/flask-security/pull/788

Issue #662, opened 2017-06-22
Email confirmation behavior


kesara:
At the moment confirmation token can be logs user in automatically.
I suggest this behavior should be, so this can be configured as required.

Other issue with the current confirmation tokens is they can be reused to login without using password as long as that token is not expired.

jirikuncar:
@kesara thank you for reporting. We should make sure that any of the tokens (confirm, reset) is not reusable. The confirmation should fail if the user has been already confirmed.

The problematic parts with automatic login are:
- https://github.com/mattupstate/flask-security/blob/develop/flask_security/views.py#L232-L234
- https://github.com/mattupstate/flask-security/blob/develop/flask_security/views.py#L294

Do you have a proposal for the configuration option `SECURITY_...LOGIN...`?

My observation is that we should not execute `login_user(user)` and logout only if the `user != current_user`.

```python
    if current_user.is_authenticated and user != current_user:
        logout_user()
```

WDYT?

kesara:
@jirikuncar I agree, it's better to logout if the `user != current_user`.
But I think it's better to make `login_user(user)` on unconfirmed accounts configurable with a configuration option like `SECURITY_CONFIRMABLE_AUTO_LOG`.
So for those who using that feature can keep using it with setting that option.

jirikuncar:
@kesara I am just trying to understand the possible situations in combination with `_security.login_without_confirmation`. IMHO we should ask user at least once for the password after registration if the confirmation is required, but it should be probably configurable as you proposed. The option should be same also for the `reset_password` use-case.

WDYT about following option `SECURITY_AUTO_LOGIN_AFTER_TOKEN = {'confirm', 'reset'}`?

kesara:
Did you meant to say `SECURITY_AUTO_LOGIN_AFTER_TOKEN = {'confirm', 'login'}` because user is supposed to enter a new password when they register? But `{'confirm', 'login'}` could be confusing.

Isn't `login_without_confirmation` only allow unconfirmed users to login? In that case I don't see this code change affecting that functionality.

I agree that it's better to ask for the password after confirmation, I only suggested that because that change could break the existing functionality. May be the for clarity and security better off removing `login_user(user)` from `confirm_email(token)`?

jirikuncar:
I would say that we remove the automatic login even if it less convenient for users.

[...]

PR #664, opened 2017-06-24
#662 Remove automatic login on email confirmation.

kesara:
Fix for #662: removes automatic login on confirmation.

### DIFF
--- a/tests/test_confirmable.py
+++ b/tests/test_confirmable.py
@@ -189,7 +189,8 @@ def test_confirmation_different_user_when_logged_in(client, get_message):
 
     response = client.get('/confirm/' + token2, follow_redirects=True)
     assert get_message('EMAIL_CONFIRMED') in response.data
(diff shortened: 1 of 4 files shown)


tests/test_confirmable.py
```diff
@@ -189,7 +189,8 @@ def test_confirmation_different_user_when_logged_in(client, get_message):
 
     response = client.get('/confirm/' + token2, follow_redirects=True)
     assert get_message('EMAIL_CONFIRMED') in response.data
-    assert b'Hello lady@lp.com' in response.data
+    assert b'Hello lady@lp.com' not in response.data
+    assert b'Hello dude@lp.com' not in response.data
```
jirikuncar: can you please add a check about the response url? I think it should be redirected to login page in this case.

jirikuncar:
@kesara shall we remove the automatic login also for password reset?

kesara:
@jirikuncar sounds good.

PR #671, opened 2017-06-29
Fix referrer loop in _get_unauthorized_view().

### LINKED ISSUE #670: _get_unauthorized_view() results in a redirect loop if the referrer is the page we just came from
`_get_unauthorized_view()` is blindly redirecting to the location in referrer; if we just came from there, this will result in a redirect loop.

nfvs:
_get_unauthorized_view() is blindly trying to redirect to the referrer,
without checking if we came from there in the first place. This change
avoids the loops that result from that, by redirecting to `/` if we just
came from the referrer.

A unit test is also provided.

Without this change, the end result is that the browser enters into a
redirect loop, until eventually it goes to the correct place (I haven't
figured out how many; in production we've seen 10s, but in the provided
unit test it only happened 3 times). It also results in multiple
flash messages being printed once the final page is loaded.

Fixes #670

### DIFF
--- a/flask_security/decorators.py
+++ b/flask_security/decorators.py
@@ -53,7 +53,12 @@ def _get_unauthorized_view():
             except BuildError:
                 view = None
         utils.do_flash(*utils.get_message('UNAUTHORIZED'))
-        return redirect(view or request.referrer or '/')
+        redirect_to = '/'
+        if (request.referrer and
+                not request.referrer.split('?')[0].endswith(request.path)):
+            redirect_to = request.referrer
+
+        return redirect(view or redirect_to)
     abort(403)
 
 
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -133,6 +133,25 @@ def test_unauthorized_access(client, get_message):
     assert get_message('UNAUTHORIZED') in response.data
 
 
+def test_unauthorized_access_with_referrer(client, get_message):
+    authenticate(client, 'joe@lp.com')
+    response = client.get('/admin', headers={'referer': '/admin'})
+    assert response.headers['Location'] != '/admin'
+    client.get(response.headers['Location'])
+
+    response = client.get(
+        '/admin?a=b',
+        headers={'referer': 'http://localhost/admin?x=y'}
+    )
+    assert response.headers['Location'] == 'http://localhost/'
+    client.get(response.headers['Location'])
+
+    response = client.get('/admin',
+                          headers={'referer': '/admin'},
+                          follow_redirects=True)
+    assert response.data.count(get_message('UNAUTHORIZED')) == 1
+
+
 def test_roles_accepted(client):
     for user in ("matt@lp.com", "joe@lp.com"):
         authenticate(client, user)


flask_security/decorators.py
```diff
@@ -53,7 +53,11 @@ def _get_unauthorized_view():
             except BuildError:
                 view = None
         utils.do_flash(*utils.get_message('UNAUTHORIZED'))
-        return redirect(view or request.referrer or '/')
+        redirect_to = '/'
+        if request.referrer and not request.referrer.endswith(request.path):
```
jirikuncar: I agree with the idea of checking the `request.referrer` and `request.path` overlap. I am not sure if it work well if the referrer contains a query arguments.

a. we should check overlap with `request.full_path`
b. compare parsed and normalized URLs (with/without host, relative/absolute path) ...

nfvs: a. disagree, since a referer to the same URL with different query arguments will not match, but still trigger a redirect loop, which is what I want to fix;
b. it's already doing that by comparing with `.endswith()`, it will work with and without hostname; in fact, the included unit test already tests both

jirikuncar: I hope it will be clearer from the code:
```diff
diff --git a/flask_security/decorators.py b/flask_security/decorators.py
index e1799a2..c1f00f0 100644
--- a/flask_security/decorators.py
+++ b/flask_security/decorators.py
@@ -54,7 +54,8 @@ def _get_unauthorized_view():
                 view = None
         utils.do_flash(*utils.get_message('UNAUTHORIZED'))
         redirect_to = '/'
-        if request.referrer and not request.referrer.endswith(request.path):
+        if request.referrer and not request.referrer.endswith(
+                request.full_path.rstrip('?')):
             redirect_to = request.referrer

         return redirect(view or redirect_to)
diff --git a/tests/test_common.py b/tests/test_common.py
index 4e76a33..2111dd3 100644
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -136,12 +136,12 @@ def test_unauthorized_access(client, get_message):
 def test_unauthorized_access_with_referrer(client, get_message):
     authenticate(client, 'joe@lp.com')
     response = client.get('/admin', headers={'referer': '/admin'})
-    assert response.headers['Location'] != '/admin'
+    assert response.headers['Location'][-6:] != '/admin'
     client.get(response.headers['Location'])

     response = client.get('/admin',
                           headers={'referer': 'http://localhost/admin'})
-    assert response.headers['Location'] != '/admin'
+    assert response.headers['Location'][-6:] != '/admin'
     client.get(response.headers['Location'])

     response = client.get('/admin',
@@ -149,6 +149,14 @@ def test_unauthorized_access_with_referrer(client, get_message):
                           follow_redirects=True)
     assert response.data.count(get_message('UNAUTHORIZED')) == 1

+    response = client.get('/admin?x=1',
+                          headers={'referer': '/admin?x=1'},
+                          follow_redirects=True)
+    assert response.data.count(get_message('UNAUTHORIZED')) == 1
+
+    response = client.get('/admin?x=1',
+                          headers={'referer': '/admin?x=0'})
+    assert response.headers['Location'][-10:] == '/admin?x=0'

 def test_roles_accepted(client):
     for user in ("matt@lp.com", "joe@lp.com"):
```

nfvs: > `request.full_path.rstrip('?')`

I'm sorry, but how is `request.full_path` without the query string different from `request.path`?

http://flask-.readthedocs.io/en/latest/api/#flask.Request.path (look at the example provided)

I can see a problem if referrer itself has a query string, but that code doesn't seem to solve it.

jirikuncar: @nfvs you can check the results of the tests if you revert `flask_security/decorators.py`. I am usually checking the official docs on http://flask.pocoo.org/docs/0.12/api/#flask.Request

nfvs: If taking `path.full_path` and removing the query string is the same as `request.path`, of course it won't change the meaning of the tests...

nfvs:
Updated to properly handle query strings: if the current page and the referrer are the same, but with different query strings, it will still avoid the redirect loop.

nfvs:
@jirikuncar any updates/comments?

nfvs:
Thanks a lot :-)

Issue #357, opened 2015-01-09
Avoid Timing Attack


cancan101:
Using `flask-security`, how would you prevent [issues like this](http://cwe.mitre.org/data/definitions/208.html): where the amount of time to check password for a valid and invalid user are the same?

mattupstate:
@cancan101 `passlib`, the underlying password encryption library, uses a [safe string comparison](https://code.google.com/p/passlib/source/browse/passlib/utils/__init__.py#250) function to prevent timing attacks.

cancan101:
That is not the issue that I am talking about. The issue is that the hashing never happens for a non existing user. See also this: http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2004-1602. 

I am assuming that `USER_DOES_NOT_EXIST` and `INVALID_PASSWORD` are set to the same message.

taion:
I think the bigger issue vis-à-vis issues like this CVE is that the default configuration explicitly informs the user whether the login failed because of unknown username or because of incorrect password.

Specifically the issue in question is a CWE-203 issue, per the supplied example 1: http://cwe.mitre.org/data/definitions/203.html

I do not think that explicitly falling afoul of the CWE is a reasonable default.

justinwp:
As linked in #227, https://kev.inburke.com/kevin/invalid-username-or-password-useless/ it is generally possible to determine the existence of a user in registration already. 
More here: http://security.stackexchange.com/questions/40694/disclose-to-user-if-account-exists

Flask Security is not setup with a workflow to handle not disclosing the username. Neither is Facebook, Amazon, Gmail etc...

taion:
There is an argument that, even from a UX perspective, distinguishing between user name and password validity is not ideal:

http://ux.stackexchange.com/questions/13516/how-to-tell-the-user-his-login-credentials-are-incorrect

Regardless of whether it's possible to get user names on Facebook/Amazon/Gmail, it remains the case that all 3 services do only have a single "invalid username/password" error message. While Kevin Burke makes some interesting points, note that Twilio, where he was at previously, also has a single "The email address or password you entered is incorrect" error on their login page.

I don't think it should be impossible for the user of a package to notify the user of whether it was the username or the password that was incorrect, but I think there's an overwhelming amount of evidence that standard best practice, and hence a reasonable default, is to not be specific here.

mattupstate:
[...]

PR #676, opened 2017-07-10
Fix timing attack on login form

sh4r3m4n:
As detailed in #357 the time it takes to process a login request is
considerably less if the user specified doesn't exist than if the
password is incorrect. This can be used as a user enumeration attack,
even if the login error messages were customized to avoid this.

I fixed it by increasing the response time of a non-existing user
request by hashing the given password anyway (if using good
password hashing algorithm this is what takes a relatively
large amount of time and makes the attack possibly)

Without the fix:
```bash
$ time curl 'http://localhost:5000/login' --data 'next=&email=not%40existing.com&password=12312312312&next=&submit=Login' -s> /dev/null
curl 'http://localhost:5000/login' --data  -s > /dev/null  0,02s user 0,00s system 43% cpu 0,037 total
$ time curl 'http://localhost:5000/login' --data 'next=&email=existing%40gmail.com&password=12312312312&next=&submit=Login' -s> /dev/null
curl 'http://localhost:5000/login' --data  -s > /dev/null  0,01s user 0,01s system 4% cpu 0,401 total
```

With the fix:
```bash
$ time curl 'http://localhost:5000/login' --data 'next=&email=not%40existing.com&password=12312312312&next=&submit=Login' -s> /dev/null
curl 'http://localhost:5000/login' --data  -s > /dev/null  0,01s user 0,01s system 4% cpu 0,402 total
$ time curl 'http://localhost:5000/login' --data 'next=&email=existing%40gmail.com&password=12312312312&next=&submit=Login' -s> /dev/null
curl 'http://localhost:5000/login' --data  -s > /dev/null  0,01s user 0,01s system 4% cpu 0,401 total
```

### DIFF
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -20,9 +20,9 @@ from wtforms import BooleanField, Field, HiddenField, PasswordField, \
     StringField, SubmitField, ValidationError, validators
 
 from .confirmable import requires_confirmation
-from .utils import _, _datastore, config_value, get_message, \
-    localize_callback, url_for_security, validate_redirect_url, \
-    verify_and_update_password, encrypt_password
+from .utils import _, _datastore, config_value, encrypt_password, \
+    get_message, localize_callback, url_for_security, validate_redirect_url, \
+    verify_and_update_password
 
 lazy_gettext = make_lazy_gettext(lambda: localize_callback)
 
--- a/flask_security/forms.py
+++ b/flask_security/forms.py
@@ -22,7 +22,7 @@ from wtforms import BooleanField, Field, HiddenField, PasswordField, \
 from .confirmable import requires_confirmation
 from .utils import _, _datastore, config_value, get_message, \
     localize_callback, url_for_security, validate_redirect_url, \
-    verify_and_update_password
+    verify_and_update_password, encrypt_password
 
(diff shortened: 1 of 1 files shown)


flask_security/forms.py
```diff
@@ -22,7 +22,7 @@
 from .confirmable import requires_confirmation
 from .utils import _, _datastore, config_value, get_message, \
     localize_callback, url_for_security, validate_redirect_url, \
-    verify_and_update_password
+    verify_and_update_password, encrypt_password
```
jirikuncar: @sh4r3m4n can you please keep the imports sorted alphabetically? It's not strictly checked, but it would be nice to keep the imports organized. Thanks!

jirikuncar:
@sh4r3m4n the PR looks reasonable 👍  Thanks!

flask_security/forms.py
sh4r3m4n: Done! Thanks for the advice!

fuhrysteve:
Looks like there's a depreciation warning in there that suggests using `hash_password` instead - does it make sense to change this PR to use `hash_password` instead of `encrypt_password`?

sh4r3m4n:
Hi @fuhrysteve, you're right, I changed this to use hash_password instead. Thanks for the advice!

PR #679, opened 2017-07-17
Optimize DB accesses by using an SQL JOIN when retrieving a user.

nfvs:
`Datastore.find_user()` is used on every page load to verify the user in
the current session; until now, `find_user()` would trigger 2+ DB
accesses, since the `roles` relationship uses lazy loading by default.
This is useful to reduce the rows accessed when retrieving multiple
users from the DB.

However, in `find_user()` we always retrieve only one, so
it makes sense to retrieve all roles in one SQL operation instead of
always triggering 2 or more.

This would be unnecessary if the `roles` relationship was defined with
`lazy="joined"`. However, that would greatly increase the number of rows
accessed on all queries (even when querying all users), so I think it's
better to be selective and optimize only these accesses which happen on
every page load.

### DIFF
--- a/flask_security/datastore.py
+++ b/flask_security/datastore.py
@@ -234,18 +234,17 @@ class SQLAlchemyUserDatastore(SQLAlchemyDatastore, UserDatastore):
 
     def get_user(self, identifier):
         from sqlalchemy import func as alchemyFn
-        from sqlalchemy.orm import joinedload
+        user_model_query = self.user_model.query
+        if hasattr(self.user_model, 'roles'):
+            from sqlalchemy.orm import joinedload
+            user_model_query = user_model_query.options(joinedload('roles'))
+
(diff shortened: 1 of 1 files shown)


jirikuncar:
@nfvs I could see only one minor issue when `role_model` is not defined. It is not very common use case and I am not sure if we want to support it, but if I remember well someone was submitting small patches to make it work.

nfvs:
Fair point, I did think of that myself but noticed that there were some references to hardcoded `roles` in the datastore anyway (though to be fair, they seem to handle cases where it doesn't exist).

I've added an additional `hasattr()` check before setting that option.

jirikuncar:
@nfvs perfect! Many thanks 👏

nfvs:
Let me know once you'd like to merge, I can squash the commits.

PR #689, opened 2017-08-05
cleaner solution found in pull request #675

Morabaraba:
Not sure how to do a test case as asked for in #675. If it pass the build can it be merged?

The solution @sbagan suggested is cleaner and more flexible compared to the atm merged solution #682.

Thanks,


jirikuncar:
merged in #675

PR #691, opened 2017-08-09
Documentation/Quickstart: Fix SQLAlchemy app example for missing render method and adding salt

KshitijKarthick:
Details:
* Fix exception raised when 'SECURITY_PASSWORD_SALT' is not set, as 'SECURITY_PASSWORD_HASH' [bcrypt] requires it which is default.
* render method is not found, replaced the same with render_template_string.
* render_template_string is utilised to show an example of usage of current user retaining the previous use case.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -108,7 +108,6 @@ and models.py. You can also do the models a folder and spread your tables there.
 
 - app.py ::
 
-    import uuid
     from flask import Flask, render_template_string
     from flask_security import Security, current_user, login_required, \
          SQLAlchemySessionUserDatastore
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -118,6 +119,8 @@ and models.py. You can also do the models a folder and spread your tables there.
     app = Flask(__name__)
     app.config['DEBUG'] = True
     app.config['SECRET_KEY'] = 'super-secret'
+    # Bcrypt is set as default SECURITY_PASSWORD_HASH, which requires a salt
+    app.config['SECURITY_PASSWORD_SALT'] = uuid.uuid4().hex
```
jirikuncar: I am not sure if we want to generate in examples. I have seen people adding the same code to their repo and then complaining that the site is not working after restart.

PR #695, opened 2017-08-16
Making sure we pass the sender address in send_mail() as a string, no…

### LINKED ISSUE #693: Cannot send registration email
When I configure the app to send registration emails, registration fails with:

ValueError: too many values to unpack

See https://github.com/mattupstate/flask-mail/issues/148

I traced back the error to this line:
https://github.com/mattupstate/flask-security/blob/develop/flask_security/utils.py#L387
the _security var is a LocalProxy

cc @filipmihal

ochronus:
…t a LocalProxy instance. Fixes #693 https://github.com/mattupstate/flask-security/issues/693

### DIFF
--- a/flask_security/utils.py
+++ b/flask_security/utils.py
@@ -384,7 +384,7 @@ def send_mail(subject, recipient, template, **context):
     context.update(_security._run_ctx_processor('mail'))
 
     msg = Message(subject,
-                  sender=_security.email_sender,
+                  sender=str(_security.email_sender),
                   recipients=[recipient])
 
     ctx = ('security/email', template)


spiropan:
Any updates or reasons why this hasn't been merged yet to master? Apparently it requires at least one approved review which I think Chamaeleon provided above? Thanks

spiropan:
Thanks I see this was merged into mattupstate:develop, but not the master branch. Does this mean when I use pip install flask-security it won't include the merged changes? Thanks,

mianos:
Correct, unless you use a -e git install with pip, flask-mail won't work if you use the app.config to set the default from MAIL_DEFAULT_SENDER or EMAIL_SENDER.

spiropan:
Thanks, I can now set set app.config['SECURITY_SEND_REGISTER_EMAIL'] = True
and sending email works now after running

# pip install --upgrade git+
https://github.com/mattupstate/flask-security.git@develop

I am using python2.7-onbuild Docker image so I guess I can add the above
line as a CMD for now.

Any idea on when this change would get merged into the master branch?

Not sure if this is the right thread but when I take it one step further
and set app.config['SECURITY_CONFIRMABLE']=True my registration page no
longer shows the extra password confirmation entry form.

I tracked down issue #601 and #580 and, unless I'm missing something, if we
want both a confirmation email and password confirmation entry form we just
have to set the below flags:

app.config['SECURITY_CONFIRMABLE']=True
app.config['SECURITY_RETYPABLE']=True

However I'm still only getting one password entry form on the registration
page. Any ideas how I can easily get both features?

Thanks again

Chamaeleon-:
@spiropan I just tested the master plus this PR and I can not reproduce your problem. I am using a registration page with email confirmation and an extra password confirmation field. 

I have only set app.config['SECURITY_CONFIRMABLE']=True and  I am not using SECURITY_RETYPABLE, but in my case it works. I think your problem is more related to #601 and #580 and should be posted there.

leonprou:
I have this issue with Python 3.5 and the fix helped.

You consider to release a new tag for this?

maximedb:
When will this be released in the master branch ?

synweap15:
Any chances to get this into a new tag? :)

PR #697, opened 2017-08-17
Add base template to security templates

grihabor:
Created ``security/base.html`` (Used edited version of [Flask-Bootstrap/base.html](https://github.com/mbr/flask-bootstrap/blob/master/flask_bootstrap/templates/bootstrap/base.html))
All security templates now extend ``security/base.html``

### DIFF
--- a/flask_security/templates/security/base.html
+++ b/flask_security/templates/security/base.html
@@ -1,30 +1,30 @@
 {% block doc -%}
-    <!DOCTYPE html>
-    <html{% block html_attribs %}{% endblock html_attribs %}>
-    {%- block html %}
-        <head>
-            {%- block head %}
-                <title>{% block title %}{{ title|default }}{% endblock title %}</title>
+<!DOCTYPE html>
+<html{% block html_attribs %}{% endblock html_attribs %}>
+  {%- block html %}
+  <head>
+    {%- block head %}
+    <title>{% block title %}{{ title|default }}{% endblock title %}</title>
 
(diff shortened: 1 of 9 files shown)


williamhatcher:
That's useful!

grihabor:
@muttupstate

flask_security/templates/security/change_password.html
```diff
@@ -1,11 +1,14 @@
+{% extends "security/base.html" %}
 {% from "security/_macros.html" import render_field_with_errors, render_field %}
-{% include "security/_messages.html" %}
```
jirikuncar: can you please keep the original style of indentation?

flask_security/templates/security/base.html
```diff
@@ -0,0 +1,30 @@
+{% block doc -%}
+    <!DOCTYPE html>
```
jirikuncar: Do not indent the first level and use only 2 spaces. Thanks

flask_security/templates/security/base.html
```diff
@@ -0,0 +1,30 @@
+{% block doc -%}
```
jirikuncar, line 1: can we call it `document` or `base`? I don't insist on a changing it here if you have any reason for it.

grihabor: I think it's better to keep it the same as in Flask-Bootstrap to avoid potential integration errors.

grihabor:
@jirikuncar fixed

PR #703, opened 2017-08-30
bugfix: support application factory pattern

briancappello:
Currently, this extension cannot be used with the factory pattern, except by subclassing `Security` to override the `__init__` and `init_app` methods. The problem is two-fold: `kwargs` passed to `__init__` don't get passed on to `init_app`, and secondly, when setting the `datastore` via `__init__`, when `init_app(app)` is later called, it wipes out the `datastore`. This PR replaces #495, which appears to have stalled.

The datastore bug was introduced in #537, and was warned about by @mattupstate in a [comment](https://github.com/mattupstate/flask-security/pull/537#issuecomment-272888626), but that PR got accidentally merged anyway.

The following is currently the only method of initializing the extension:
```python
app = Flask(__name__)
datastore = SQLAlchemyUserDatastore(db, User, Role)
security = Security(app, datastore, **kwargs)
```
It remains supported, and this PR makes it possible to also initialize this way:
```python
# extensions.py
datastore = SQLAlchemyUserDatastore(db, User, Role)
security = Security(datastore=datastore, **kwargs)

# later, in create_app.py
from extensions import security
def create_app():
    app = Flask(__name__)
    security.init_app(app)
    return app
```

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -478,20 +478,15 @@ class Security(object):
         self.app = app
         self._datastore = datastore
         self._register_blueprint = register_blueprint
-        self._kwargs = dict(login_form=None,
-                            register_form=None,
-                            confirm_register_form=None,
-                            forgot_password_form=None,
-                            reset_password_form=None,
-                            change_password_form=None,
-                            send_confirmation_form=None,
-                            passwordless_login_form=None,
-                            anonymous_user=None)
-        self._kwargs.update(kwargs)
+        self._kwargs = kwargs
 
         self._state = None  # set by init_app
         if app is not None and datastore is not None:
-            self._state = self.init_app(app, datastore, **self._kwargs)
+            self._state = self.init_app(
+                app,
+                datastore,
+                register_blueprint=register_blueprint,
+                **kwargs)
 
     def init_app(self, app, datastore=None, register_blueprint=None, **kwargs):
         """Initializes the Flask-Security extension for the specified
@@ -502,11 +497,15 @@ class Security(object):
         :param register_blueprint: to register the Security blueprint or not.
         """
         self.app = app
+
         if datastore is None:
             datastore = self._datastore
+
         if register_blueprint is None:
             register_blueprint = self._register_blueprint
-        self._kwargs.update(kwargs)
+
+        for key, value in self._kwargs.items():
+            kwargs.setdefault(key, value)
 
         for key, value in _default_config.items():
             app.config.setdefault('SECURITY_' + key, value)
@@ -516,7 +515,7 @@ class Security(object):
 
         identity_loaded.connect_via(app)(_on_identity_loaded)
 
-        self._state = state = _get_state(app, datastore, **self._kwargs)
+        self._state = state = _get_state(app, datastore, **kwargs)
 
         if register_blueprint:
             app.register_blueprint(create_blueprint(state, __name__))
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -461,30 +461,52 @@ class Security(object):
 
     :param app: The application.
     :param datastore: An instance of a user datastore.
+    :param register_blueprint: to register the Security blueprint or not.
+    :param login_form: set form for the login view
+    :param register_form: set form for the register view
+    :param confirm_register_form: set form for the confirm register view
+    :param forgot_password_form: set form for the forgot password view
+    :param reset_password_form: set form for the reset password view
+    :param change_password_form: set form for the change password view
+    :param send_confirmation_form: set form for the send confirmation view
+    :param passwordless_login_form: set form for the passwordless login view
+    :param anonymous_user: class to use for anonymous user
     """
 
-    def __init__(self, app=None, datastore=None, **kwargs):
+    def __init__(self, app=None, datastore=None, register_blueprint=True,
+                 **kwargs):
         self.app = app
-        self.datastore = datastore
+        self._datastore = datastore
+        self._register_blueprint = register_blueprint
+        self._kwargs = dict(login_form=None,
+                            register_form=None,
+                            confirm_register_form=None,
+                            forgot_password_form=None,
+                            reset_password_form=None,
+                            change_password_form=None,
+                            send_confirmation_form=None,
+                            passwordless_login_form=None,
+                            anonymous_user=None)
+        self._kwargs.update(kwargs)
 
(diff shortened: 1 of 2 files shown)


tests/test_datastore.py
```diff
@@ -161,8 +161,8 @@ def test_delete_user(app, datastore):
 
 
 def test_access_datastore_from_factory(app, datastore):
-    security = Security()
-    security.init_app(app, datastore)
```
jirikuncar: for compatibility reasons this test case must be supported.

flask_security/core.py
```diff
@@ -461,30 +461,48 @@ class Security(object):
...
+    :param passwordless_login_form: set form for the passwordless login view
+    :param anonymous_user: class to use for anonymous user
     """
 
     def __init__(self, app=None, datastore=None, **kwargs):
         self.app = app
         self.datastore = datastore
-
+        self.kwargs = dict(register_blueprint=True,
+                           login_form=None,
```
jirikuncar: do you need to set all the default values to `None`?

flask_security/core.py
```diff
@@ -461,30 +461,48 @@ class Security(object):
...
+                           register_form=None,
+                           confirm_register_form=None,
+                           forgot_password_form=None,
+                           reset_password_form=None,
+                           change_password_form=None,
+                           send_confirmation_form=None,
+                           passwordless_login_form=None,
+                           anonymous_user=None)
+        self.kwargs.update(kwargs)
```
jirikuncar: `self.kwargs = kwargs`

flask_security/core.py
```diff
@@ -461,30 +461,48 @@ class Security(object):
...
-        application and datastore implentation.
+        application and datastore implementation.
 
         :param app: The application.
         :param datastore: An instance of a user datastore.
-        :param register_blueprint: to register the Security blueprint or not.
         """
         self.app = app
-        self.datastore = datastore
+        if datastore is not None:
```
jirikuncar: IMHO we should not change the global datastore.

```python
datastore = datastore or self.datastore
kwargs.update(self.kwargs)

# ...

self._state  = state = _get_state(app, datastore, **kwargs)
```

flask_security/core.py
```diff
@@ -494,37 +512,26 @@ def init_app(self, app, datastore=None, register_blueprint=True,
...
+        self._state = _get_state(app, self.datastore, **self.kwargs)
+
+        if self.kwargs['register_blueprint']:
+            app.register_blueprint(create_blueprint(self._state, __name__))
             app.context_processor(_context_processor)
 
         @app.before_first_request
         def _register_i18n():
             if '_' not in app.jinja_env.globals:
-                app.jinja_env.globals['_'] = state.i18n_domain.gettext
```
jirikuncar: keep the changes minimal by using trick in https://github.com/mattupstate/flask-security/pull/703/files#r143207783

flask_security/core.py
briancappello: If I'm reading the code right, I think so, because the keys need to be set somewhere so that `_SecurityState` can initialize its attributes: https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L410 Previously that happened in the call to `_get_state`, but now that the kwargs can be passed to either `__init__` and/or `init_app`, this seemed to me the most DRY way of doing it. Does that make sense, or perhaps I am missing something?

jirikuncar: Then I think we should move the default values to `_get_state` function.

flask_security/core.py
```diff
@@ -494,18 +511,9 @@ def init_app(self, app, datastore=None, register_blueprint=True,
...
-                           reset_password_form=reset_password_form,
-                           change_password_form=change_password_form,
-                           send_confirmation_form=send_confirmation_form,
-                           passwordless_login_form=passwordless_login_form,
-                           anonymous_user=anonymous_user)
-
-        if register_blueprint:
+        self._state = state = _get_state(app, datastore, **kwargs)
+
+        if kwargs['register_blueprint']:
```
jirikuncar: IMHO we should keep `register_blueprint` in arguments since it is not really a state (and handle it similarly to `datastore`).

flask_security/core.py
briancappello: I made this change, but there could be one possibly surprising behaviour with the kwargs:
```
kwargs.update(self.kwargs)
```
Doing it this way, kwargs passed to `__init__` would always take precedence over those passed to `init_app`. I'm not sure why somebody would want to pass an argument to the constructor and then later override it in the call to `init_app`, but, something to note.

briancappello:
@jirikuncar Once you think this is ready, LMK and I can squash the commits

briancappello:
I remember now why I had to set the kwarg defaults in `__init__` instead of the `_get_state` call:
```
        self._state = state = _get_state(app, datastore,
                                         login_form=None,
                                         register_form=None,
                                         confirm_register_form=None,
                                         forgot_password_form=None,
                                         reset_password_form=None,
                                         change_password_form=None,
                                         send_confirmation_form=None,
                                         passwordless_login_form=None,
                                         anonymous_user=None,
>                                        **kwargs)
E       TypeError: _get_state() got multiple values for keyword argument 'login_form'
```
So, I'm not really sure of a better way to do it than what I did originally. I added a test to check for this.

flask_security/core.py
```diff
@@ -461,30 +461,52 @@ class Security(object):
...
+    :param anonymous_user: class to use for anonymous user
     """
 
-    def __init__(self, app=None, datastore=None, **kwargs):
+    def __init__(self, app=None, datastore=None, register_blueprint=True,
+                 **kwargs):
         self.app = app
-        self.datastore = datastore
-
+        self._datastore = datastore
```
briancappello: I prefixed this attribute with an underscore so that both test cases pass whether datastore was passed to `__init__` or `init_app`. The issue with it being named `datastore` is that that takes precedence over calls to `Security.__getattr__()`:

```python
def test_access_datastore_from_factory(app, datastore):
    security = Security()  # sets Security.datastore to None
    security.init_app(app, datastore)  # sets Security._state['datastore'] to datastore

    assert security.datastore is not None  # fails, because this retrieves Security.datastore, not Security.__getattr__('datastore'), which would return Security._state['datastore']
```

briancappello:
@jirikuncar OK, tests are passing and post-review commits squashed

PR #709, opened 2017-09-14
Fix bug

zhumaobi:


jirikuncar:
@zhumaobi can you please rebase the branch, add explanation of a bug you are trying to fix and use english for all commit messages? Thank you!

jirikuncar:
These modification should be moved to your own package as configuration options.

PR #714, opened 2017-09-29
Make SECURITY_PASSWORD_SINGLE_HASH a list of scheme ignoring double hash

noirbizarre:
This PR makes `SECURITY_PASSWORD_SINGLE_HASH` a list of password hashing schemes that shouldn't be hashed twice allowing easier handling of legacy schemes or imported schemes.

## Use cases
- I want to import my Django passwords in my Flask-Security based app and have `bcrypt` with double hashing as default scheme, but Django only perform single hashing. (handled in default configuration)
- I add Flask-Security to default on `pbkdf2_sha256` hashed once but I want to migrate to `bcrypt` hashed twice without loosing already stored passewords. This can be handled with:
```python
'SECURITY_PASSWORD_HASH': 'bcrypt',
'SECURITY_PASSWORD_SALT': 'any string',
'SECURITY_PASSWORD_SINGLE_HASH': ['pbkdf2_sha256'],
'SECURITY_PASSWORD_SCHEMES': ['bcrypt', 'pbkdf2_sha256']
```

## Backward compatibility
This ensure backward compatibility with the actual behavior: `SECURITY_PASSWORD_SINGLE_HASH = True` disable double hashing and `SECURITY_PASSWORD_SINGLE_HASH = False` ignore double hashing for `plaintext`.

## Default
The default `SECURITY_PASSWORD_SINGLE_HASH` is a list of known scheme not working with double hashing (`django_*` schemes and `plaintext`)

## Side effect
This PR also remove the need to raise an exception if `SECURITY_PASSWORD_SALT` with `SECURITY_PASSWORD_SINGLE_HASH` is defined: having it defined does not break the security, it is simply ignored when not necessary.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -21,7 +21,7 @@ Core
 ``SECURITY_URL_PREFIX``                  Specifies the URL prefix for the
                                          Flask-Security blueprint. Defaults to
                                          ``None``.
-``SECURITY_SUBDOMAIN``                   Specifies the subdomain for the 
+``SECURITY_SUBDOMAIN``                   Specifies the subdomain for the
                                          Flask-Security blueprint. Defaults to
                                          ``None``.
 ``SECURITY_FLASH_MESSAGES``              Specifies whether or not to flash
@@ -46,6 +46,11 @@ Core
                                          ``SECURITY_PASSWORD_SALT``, and then
                                          with a random salt. May be useful for
                                          integrating with other applications.
+                                         It can also be a set of scheme that
+                                         should not be hashed twice.
+                                         Default to a list of known schemes
+                                         not working with double hashing
+                                         (`django_{digest}`, `plaintext`).
                                          Defaults to ``False``.
 ``SECURITY_HASHING_SCHEMES``             List of algorithms used for
                                          creating and validating tokens.
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -48,7 +48,17 @@ _default_config = {
     'I18N_DOMAIN': 'flask_security',
     'PASSWORD_HASH': 'bcrypt',
     'PASSWORD_SALT': None,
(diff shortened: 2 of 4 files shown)


docs/configuration.rst
```diff
@@ -46,6 +46,11 @@ Core
                                          ``SECURITY_PASSWORD_SALT``, and then
                                          with a random salt. May be useful for
                                          integrating with other applications.
+                                         It can also be a list of scheme that
```
jirikuncar: list -> set

flask_security/core.py
```diff
@@ -48,7 +48,17 @@
     'I18N_DOMAIN': 'flask_security',
     'PASSWORD_HASH': 'bcrypt',
     'PASSWORD_SALT': None,
-    'PASSWORD_SINGLE_HASH': False,
+    'PASSWORD_SINGLE_HASH': [
```
jirikuncar: I would propose to change the default from list to set.

flask_security/utils.py
```diff
@@ -447,17 +447,16 @@ def get_identity_attributes(app=None):
 
 def use_double_hash(password_hash=None):
     """Return a bool indicating whether a password should be hashed twice."""
-    single_hash = config_value('PASSWORD_SINGLE_HASH')
-    if single_hash and _security.password_salt:
-        raise RuntimeError('You may not specify a salt with '
-                           'SECURITY_PASSWORD_SINGLE_HASH')
+    # Default to plaintext for backward compatibility with
+    # SECURITY_PASSWORD_SINGLE_HASH = False
+    single_hash = config_value('PASSWORD_SINGLE_HASH') or ['plaintext']
```
jirikuncar: `{'plaintext'}`

noirbizarre:
It makes sense. Change done :+1:

PR #717, opened 2017-10-21
Allow custom login_manager to be passed in to Flask-Security

Jaza:


Jaza:
This is necessary in order to use OAuth2 (with Flask-OAuthlib) together with Flask-Security.

After applying this change, I'm able to call `security.init_app(app, datastore, login_manager=login_manager)`, where `login_manager` has a custom `request_loader` set, that first tries to load the user via OAuth, then falls back to trying the Flask-Security `request_loader`.

jirikuncar:
@Jaza can you please add docs about this change? I will merge it but it would be useful to have an example how to use Flask-Security with Flask-OAuthlib. Many thanks!

jirikuncar (approved):

Issue #678, opened 2017-07-14
AttributeError: 'NoneType' object has no attribute 'password'


amuzeev:
1.7.5 and 3.0.0
flask_security/recoverable.py:77
`if user.password:`
I get **user** as None because of this:
flask_security/utils.py:428
`
    if data:
        user = _datastore.find_user(id=data[0])
`
**_datastore** can't get user by id because user isn't exist.

PR #721, opened 2017-10-22
Bug fix for token verification on deleted users. (#678)

kesara:
* Returns invalid status when user doesn't exist. (#678)
* Added test case for token confirmation on deleted users.


kesara:
Rethinking about this, not the best way to fix this

PR #722, opened 2017-10-23
Bug fix: password recovery confirmation on deleted user

kesara:
* Test case for password recovery on deleted user.
* [#678] Bug fix: password recovery on deleted user


jirikuncar (approved):
LGTM

PR #739, opened 2017-12-31
check that user model has roles attr on user create

Jaza:


flask_security/datastore.py
```diff
@@ -125,12 +125,15 @@ def _prepare_role_modify_args(self, user, role):
...
     def _prepare_create_user_args(self, **kwargs):
         kwargs.setdefault('active', True)
-        roles = kwargs.get('roles', [])
-        for i, role in enumerate(roles):
-            rn = role.name if isinstance(role, self.role_model) else role
-            # see if the role exists
-            roles[i] = self.find_role(rn)
-        kwargs['roles'] = roles
+
+        if hasattr(self.user_model, 'roles'):
```
jirikuncar: I would rather see an exception if you try to save user with roles.

```python
if not hasattr(self.user_model, 'roles') and kwargs.get('roles'):
    raise  # custom Flask-Security exception
```

jirikuncar (changes requested):
Please add new tests. Thanks!

flask_security/datastore.py
Jaza: Ok @jirikuncar, now raising an exception as you suggested, and have added tests.

Sorry it took me a whole year to respond to your feedback. Anyway, better late than never!

jwag956:
If you are doing all this checking - would it make sense to exception if role doesn't exist - rather than just ignoring it? That would seem like a equally bad 'admin/config' error...

PR #741, opened 2018-01-03
Adds support for 'token' in email templates

pahrohfit:
Simple add of the token to be available in the email templates.

When using with an application via REST, being able to integrate with the device native app, using something like Branch.io.  This will allow a click via email to open either the native app, app store, or web site, to perform the function using the token from the request.


jirikuncar (changes requested):
Can you please add a new test to `tests/test_context_processors.py` to make sure that the token is correctly passed to all mail templates? Thanks!

PR #749, opened 2018-01-30
allow append custom salt to make sign stronger.

9nix00:
Although use SECURITY_XXX_SALT is safe, in most case. but if someone get server config, it will be dangerous.
Append a custom salt by user can make sign stronger.


jirikuncar:
@wangwenpei can you please explain how is it going to be used?

jirikuncar:
You can override any serialiser with custom class returning different salt.

If you still want to merge this changes please re-open the PR with tests and documentation. Thanks!

Issue #740, opened 2018-01-01
Slowness


BillSchumacher:
All Flask-Security related functions take ~20-30 seconds to complete. It did not use to be so slow. I thought maybe it was the emails but I moved those to a celery task, the slowness persists. I implemented other logic that uses email and it is instant.

Not sure why but this library has a huge delay for some reason.

BillSchumacher:
I decided to implement the logic for user login using the same forms, db lookup and password verification. 

Surprisingly the form validation and password hashing both take ~10 seconds.

('before validation', '2017-12-31 23:52:13.046964')
('after validation', '2017-12-31 23:52:22.855227')
('user lookup', '2017-12-31 23:52:22.855389')
('after user lookup', '2017-12-31 23:52:22.870330')
('password hash', '2017-12-31 23:52:22.870482')
('after password hash', '2017-12-31 23:52:31.045613')
('after auth', '2017-12-31 23:52:31.045836')
('after login', '2017-12-31 23:52:31.059585')

mafrosis:
https://github.com/mattupstate/flask-security/issues/731 ?

BillSchumacher:
I see that the hashing is done in the form validation, I'll have to take a look to see what else is happening to add ~20 seconds to the standard Flask-Security login. Even 9 seconds is too slow.

('before validation', '2018-01-01 00:07:56.931346')
('after validation', '2018-01-01 00:08:03.957584')
('after login', '2018-01-01 00:08:03.971493')

BillSchumacher:
So this is interesting, if I don't use the Flask-Security LoginForm and call verify_password like I was previously the issue is resolved? The same function call took 10 seconds previously.

('before validation', '2018-01-01 00:51:16.683555')
('after validation', '2018-01-01 00:51:16.684854')
('user lookup', '2018-01-01 00:51:16.684928')
('after user lookup', '2018-01-01 00:51:16.689784')
('password hash', '2018-01-01 00:51:16.689911')
('after password hash', '2018-01-01 00:51:16.691094')
('after login', '2018-01-01 00:51:16.702708')

BillSchumacher:
Not sure why but verify_password is slowed down significantly by running the forms verify_and_update_password prior to calling verify_password, verify_update_and_password is still ~10 seconds with the custom form.

BillSchumacher:
verify_and_update timings:

('verify_and_update initial', '2018-01-01 01:04:22.480948')
[...]

Issue #731, opened 2017-11-07
Slow token authentication


KPilnacek:
Hello guys,

at my current job, I happened to be a part of a backend team, which is creating an API.
The API should be then served to javascript application and needs to be quite fast (100 ms or so).
However, it is not.
After some profiling, we figured out that it is the token authentication in `Flask-security`, which is holding us back (pleas see the MWE). 

### MWE

```python
import flask
from flask_security import Security, SQLAlchemyUserDatastore, UserMixin, RoleMixin, auth_required
from flask_sqlalchemy import SQLAlchemy

app = flask.Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:////tmp/database.sqlite3'
app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
app.config['WTF_CSRF_ENABLED'] = False
app.config['SECURITY_TOKEN_AUTHENTICATION_HEADER'] = 'Authorization'
app.config['SECURITY_PASSWORD_HASH'] = 'pbkdf2_sha512'
app.config['SECURITY_PASSWORD_SALT'] = b'secret'
app.config['SECRET_KEY'] = "super_secret"

db = SQLAlchemy(app)

roles_users = db.Table('roles_users',
                       db.Column('user_id', db.Integer(), db.ForeignKey('user.id')),
                       db.Column('role_id', db.Integer(), db.ForeignKey('role.id')))


class Role(db.Model, RoleMixin):
    id = db.Column(db.Integer(), primary_key=True)
    name = db.Column(db.String(80), unique=True)
    description = db.Column(db.String(255))


class User(db.Model, UserMixin):
    id = db.Column(db.Integer, primary_key=True)
    email = db.Column(db.String(255), unique=True)
    password = db.Column(db.String(255))
    active = db.Column(db.Boolean())
    confirmed_at = db.Column(db.DateTime())
    roles = db.relationship('Role', secondary=roles_users,
                            backref=db.backref('users', lazy='dynamic'))


user_datastore = SQLAlchemyUserDatastore(db, User, Role)

# Setup Flask-Security
security = Security(app, user_datastore)

db.drop_all()
db.create_all()

admin_role = Role(**{'name': 'admin', 'description': 'Admin role'})
db.session.add(admin_role)
db.session.commit()

user_datastore.create_user(email='test@example.com', password='test', active=True, roles=[Role.query.first()])
db.session.commit()


@app.route('/')
[...]

PR #753, opened 2018-02-09
Remove password hashing for each token login

LOUISLCE:
Hi,
The token request loader was critically slowing down login process according to #731 and #740 .

This remove hashing the user password at every request, which is pretty useless given that
the token is already signed by the serializer (and as such is hard to fake).

Thanks!

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -249,7 +249,7 @@ def _request_loader(request):
         user = _security.datastore.find_user(id=data[0])
         if user:
             return user
-    except Exception as e:
+    except:
         pass
     return _security.login_manager.anonymous_user()
 
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -30,7 +30,7 @@ from .forms import ChangePasswordForm, ConfirmRegisterForm, \
     ResetPasswordForm, SendConfirmationForm
 from .utils import config_value as cv
 from .utils import _, get_config, hash_data, localize_callback, string_types, \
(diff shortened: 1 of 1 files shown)


flask_security/core.py
```diff
@@ -247,7 +247,7 @@ def _request_loader(request):
         data = _security.remember_token_serializer.loads(
             token, max_age=_security.token_max_age)
         user = _security.datastore.find_user(id=data[0])
-        if user and verify_hash(data[1], user.password):
```
jirikuncar, line 250: I think the reason to call `verify_hash` was to make sure that the token is invalidated once a user changes her password. Can you propose any solution that allows users to invalidate the remember token?

jirikuncar (changes requested):

briancappello:
just thinking out loud here, i wonder if it could work to store verified tokens in the database, and whenever the user changes their password, clear out the cache. so whenever `User.get_auth_token()` gets called, it adds the token to the cache on the model, unless max age is None and a token already exists, in which case no need to generate a new one. and in the request loader, we're already doing a database lookup anyway, so any extra work should be minimal compared to the time saved in the most common case (ie that the hash has already been verified). it seems to me the most brittle part is the fact that the user model can't commit itself, and thus any code that calls `get_auth_token` needs to remember to commit the user instance (altho even in the case that that doesn't happen, the worst that happens would be a fallback to the current behaviour, ie verifying the hash every time)

jwag956:
Please see some ideas in: #771

PR #758, opened 2018-02-21
Dynamic translation and Post view configuration changes

kantorii:
This request includes the following changes:

    Prevent translated messages from being cached, so that language can be changed dynamically.
    Added configuration for SECURITY_POST_FORGOT_VIEW
    (This specifies the page to transition to after sending instructions to reset password.)
    Added configuration for POST_SEND_CONFIRMATION_VIEW
    Added many translations for Japanese.
    Renamed translation file for nl_NL, because it is likely wrong.

Please let me know if you want to remove some changes.
Of course, you are free to remove them yourself.


kantorii:
Added another commit:

    Don't flash message if there was an error sending welcome email.

flask_security/forms.py
```diff
@@ -44,8 +44,10 @@
 
 class ValidatorMixin(object):
     def __call__(self, form, field):
+        # This is a fix that avoids the API being changed.
         if self.message and self.message.isupper():
-            self.message = get_message(self.message)[0]
+            self.message_key = self.message
+        self.message = get_message(self.message_key)[0]
```
jirikuncar: What if the `self.message_key` is not set?

Can we change the Validator mixin to something around this lines?
```python
class ValidatorMixin(object):
   
    @property
    def message(self):
        if self._templated:
            return get_message(self._message)[0]
        return self._message

    @message.setter
    def message(self, value):
        self._templated = value and value.isupper()
        self._message = value
```

jirikuncar (changes requested):
Can you send a separate PR with the jp translations? Thanks

flask_security/forms.py
kantorii: Why did I miss that? :)

I'm ok with properties.
Do you want me to put it into this Pull Request?

jirikuncar: @kantorii yes, you can update this PR.

flask_security/registerable.py
```diff
@@ -40,4 +39,7 @@ def register_user(**kwargs):
                             'welcome', user=user,
                             confirmation_link=confirmation_link)
 
+        if _security.confirmable:
```
jirikuncar, line 42: @kantorii can you please clarify why do we need to move the `do_flash` call to the separate block?

kantorii: Yes. If there is an exception within user_registered.send(), do_flash() will not be executed.
I believe it should be that way, because, if I remember correctly, the flash says an email was sent.

flask_security/forms.py
kantorii: @jirikuncar Thanks for the wait. The suggested change is on the branch now.

jirikuncar (commented):
@kantorii it looks good now. Can you rebase and remove the commit with translations?

PR #760, opened 2018-02-22
Added config option 'SECURITY_USER_ACTIVE_BY_DEFAULT'

eliaperantoni:
You can use it to specify whether or not users should be active by default, useful if you want registered users to be reviewed and activated manually.

Let's say you have a small company website and you want your employees to register with their email and passwords but you want to be able to activate them manually so that you can control who gets accepted, you could do it by setting `SECURITY_USER_ACTIVE_BY_DEFAULT` to `False` like this:

```
app.config['SECURITY_USER_ACTIVE_BY_DEFAULT'] = False
```

You could easily review and accept (or reject) registration attempts with a tool like Flask-Admin

### DIFF
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -136,6 +136,7 @@ _default_config = {
     ],
     'DEPRECATED_HASHING_SCHEMES': ['hex_md5'],
     'DATETIME_FACTORY': datetime.utcnow,
+    'USER_ACTIVE_BY_DEFAULT': True,
 }
 
 #: Default Flask-Security messages
--- a/flask_security/core.py
+++ b/flask_security/core.py
@@ -136,7 +136,7 @@ _default_config = {
     ],
     'DEPRECATED_HASHING_SCHEMES': ['hex_md5'],
     'DATETIME_FACTORY': datetime.utcnow,
-    'USER_ACTIVE_BY_DEFAULT': True,
+    'MANUAL_USER_ACTIVATION': False,
 }
 
 #: Default Flask-Security messages
--- a/flask_security/datastore.py
+++ b/flask_security/datastore.py
@@ -124,7 +124,7 @@ class UserDatastore(object):
         return user, role
 
     def _prepare_create_user_args(self, **kwargs):
-        kwargs.setdefault('active', config_value('USER_ACTIVE_BY_DEFAULT'))
+        kwargs.setdefault('active', not config_value('MANUAL_USER_ACTIVATION'))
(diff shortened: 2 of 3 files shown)


flask_security/datastore.py
```diff
@@ -124,7 +124,7 @@ def _prepare_role_modify_args(self, user, role):
         return user, role
 
     def _prepare_create_user_args(self, **kwargs):
-        kwargs.setdefault('active', True)
```
jirikuncar, line 127: @eliaperantoni could we simply remove the line and rely on default in user defined datastore?

flask_security/core.py
```diff
@@ -136,6 +136,7 @@
     ],
     'DEPRECATED_HASHING_SCHEMES': ['hex_md5'],
     'DATETIME_FACTORY': datetime.utcnow,
+    'USER_ACTIVE_BY_DEFAULT': True,
```
jirikuncar: Can you please move it close to `USER_IDENTITY_ATTRIBUTES` option?

jirikuncar (changes requested):
If we need new configuration option, please include it in the documentation.

flask_security/core.py
eliaperantoni: Doesn't `USER_IDENTITY_ATTRIBUTES` specify what user attributes uniquely identify it?

If so, I didn't meant this thing to be a user-specific attribute but rather an application-wise configuration that determines whether or not new users are active by default therefore being able to login without being reviewed. That's why I made `USER_ACTIVE_BY_DEFAULT` a configuration value flask-security wise.

Thing of this way: you can set `USER_ACTIVE_BY_DEFAULT` to True and you'll have a standard registration, as soon as you register and confirm your email you can login.
On the other hand, if you set `USER_ACTIVE_BY_DEFAULT` to False, after you register you will have to wait for administrators to approve your account  before  being able to log in.

It's basically designed for restricted-access websites.

flask_security/datastore.py
eliaperantoni: By having this line here we're able to make users active or not by default but we can also override this setting for some particular users if needed.

flask_security/core.py
jirikuncar: It sounds good. I was a bit confused by the name. Can you add the documentation and consider changing the name to something around:

- `MANUAL_USER_ACTIVATION`
- `RESTRICT_REGISTRATION`
...

eliaperantoni: Sure! Check out the new commit

Issue #348, opened 2014-12-02
Brute force protection on login failure


mafrosis:
I don't believe there is anything in Flask-Security that provides IP-based timing delays to prevent brute force login attempts. Please correct me if I'm wrong!

If it's missing, I will implement it in the coming months.

nfvs:
I'm not sure doing this in Flask is a good idea, since you'd be wasting resources that would be best used to process legitimate requests.

This is something that's best done in the web server itself. If you're using Nginx, this can easily be accomplished by using the "ngx_http_limit_req_module" module: http://nginx.org/en/docs/http/ngx_http_limit_req_module.html

ldupuy:
Yes, but a recaptcha after 3 or 5 errors could be great, i think.

jonafato:
I believe this is related to (but not necessarily a duplicate of) #161.

cbronazc:
+1. The nginx or flask-limiter solutions are aimed at rate-limiting a particular IP, but that doesn't address an attacker going after one specific user. I would prefer either to lock the account after *N* failed attempts and force the user to reset via email, or lock account for *N* minutes. Here is a good [example implementation in Ruby.](http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Lockable)

rakeshmukundan:
I am looking to add "lock" support for a particular user account after N failed login. Anyone else is working on this? if else would love to do it

jirikuncar:
@rakeshmukundan you can try to propose a PR. See my other comment with possible solution using custom login form https://github.com/mattupstate/flask-security/issues/161#issuecomment-310921347. You can also disable account from the form and send them a notification via email.

Issue #161, opened 2013-09-09
Login failure throttle


mattupstate:
Do something if user fails to login X amount of times

ldupuy:
yes, but you can add the recaptcha after 3 or 5 errors.
I think it could protect the application without scare the users

amanvir:
Anybody picking this up?

rakeshmukundan:
I would like to pick this up. Anyone else have ideas?

jirikuncar:
@rakeshmukundan feel free to pick this one up. I would start by custom login form which would count every invalid validation and constructor could add recaptcha field when needed.

rakeshmukundan:
Just picking this up again. I did make a custom form which can do this. However I feel better option is to keep this functionality outside of flask-security and provide a incorrect password entered signal ? So then the application logic can decide what to do.

asif-mahmud:
besides of added recaptcha support login view should log the failed attempts with IP address if TRACKABLE flag is True. This will help using tools like fail2ban to ban the failed ip addresses for some time or permanently.

PR #764, opened 2018-03-06
Add optional recaptcha brute force protection

gnordens:
Hi,
This PR adds an optional recaptcha brute force protection to the login form (default is disabled).

Login attempts and last login attempt are tracked on each user and require two extra columns:
`last_login_attempt_at = db.Column(db.DateTime, default=db.func.current_timestamp(), nullable=False) 
  login_attempts = db.Column(db.Integer, default=0)`
A user is required to enter a recaptcha after security_recaptcha_max_attempts is reached. 
Login attempts are reset for a user after security_recaptcha_timeout is reached.

Related:
https://github.com/mattupstate/flask-security/issues/348
https://github.com/mattupstate/flask-security/issues/161


Avamander:
There already is optional recaptcha protection ability in flask-security. 
```python
from flask_wtf import RecaptchaField

class ExtendedRegistrerForm(forms.RegisterForm):
    recaptcha = RecaptchaField("Captcha")

security = Security(app, user_datastore,
                    confirm_register_form=ExtendedRegistrerForm)
```

You can override every form like this.

gnordens:
@Avamander 
I might be misunderstanding you (or missing something in the docs), but in your example aren't you simply overriding a form to include a recaptcha element? This PR instead looks to add a conditional recaptcha based on consecutive logins for a user rather than always displaying one. This is to reduce annoyance for the user trying to login while also adding some protection against brute force attacks. A similar feature has been request in the issues mentioned in the OP.

Avamander:
@gnordens Nothing stops you from counting login attempts in a custom user field inside the custom form and displaying the captcha based on that. It's almost what you're doing with this PR, the primary reason I commented was that I personally can't see a big improvement over the override method and your PR, as it doesn't touch the forms that would actually really require captchas (PW reset, register for ex.). What I'm trying to say that this is only half of the solution and just including this might give a false sense of security and every form should be sufficiently protected that f-security has.

jirikuncar:
This looks like a change that should be handled per instance basis. Personally, I would rather store the login attempts in Redis rather than in SQL database.

jirikuncar (changes requested):
Can you rewrite it as a tutorial that can be shown in docs?

flask_security/utils.py
```diff
@@ -515,3 +515,24 @@ def _on(app, **data):
...
+
+    User's login attempts are reset if they have not attempted to
+    login again within SECURITY_RECAPTCHA_TIMEOUT.
+
+    """
+    if datetime.utcnow() > user.last_login_attempt_at \
+            + timedelta(milliseconds=_security.recaptcha_timeout):
+        user.login_attempts = 0
+
+    user.login_attempts = user.login_attempts + 1 if user.login_attempts else 1
```
jirikuncar, line 536: The counter might not be very reliable for parallel attempts.

jirikuncar:
I am going to close the PR mainly for following reasons:

1. The ReCaptcha field can be already added to the custom forms;
2. The login counter is not very reliable in the current implementation;
3. Lack of tests.

If you find that 2. needs to be implemented in this package please send a separate PR.

PR #785, opened 2018-07-01
Enable token authentication for change_password()

jirikuncar:
Two changes:
1. Enable changing password by providing token
2. Provide new token in response to successful password change

---
closes #479

### DIFF
--- a/flask_security/views.py
+++ b/flask_security/views.py
@@ -303,7 +303,7 @@ def reset_password(token):
     )
 
 
-@auth_required('token', 'session')
+@auth_required('session', 'token', 'basic')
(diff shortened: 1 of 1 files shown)


jwag956:
A couple subtle behavior changes that this exposes - possibly we should address those as well.

@login_required has some additional features:
a) it allows OPTIONS method (for CORS)
b) flask-login supports a `LOGIN_DISABLED` config variable (for testing).

This change alters those 2. My feeling is that it would be better to simply teach auth_required() to support the OPTIONS feature - not sure LOGIN_DISABLED is important?

Issue #781, opened 2018-06-28
RuntimeError: The configuration value `SECURITY_PASSWORD_SALT` must not be None when the value of `SECURITY_PASSWORD_HASH` is set to "bcrypt"


fmerges:
Previously the default was plaintext, nowadays it's bcrypt, but the usage example on the website does not define the SECURITY_PASSWORD_SALT.

johnroper100:
Same issue here

PR #791, opened 2018-07-13
Fix documentation for #781

fmerges:
This fixes the documentation. But, on the same topic, currently the extension doesn't really support plaintext anymore, and I guess it's not encouraged per se:

https://github.com/mattupstate/flask-security/blob/89198288bc416a7921b4973d8e897993942bb428/flask_security/utils.py#L120-L127

So why not eliminate it completely? Additionally, the transparent update of plaintext passwords to the configured hash is also something arguable (backward compatibility?)

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -40,10 +40,8 @@ Core
                                          ``bcrypt``, ``sha512_crypt``, or
                                          ``pbkdf2_sha512``. Defaults to
                                          ``bcrypt``.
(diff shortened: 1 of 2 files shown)


jirikuncar (approved):

jirikuncar:
The plaintext passwords should still be possible: https://github.com/mattupstate/flask-security/blob/89198288bc416a7921b4973d8e897993942bb428/flask_security/utils.py#L455

PR #796, opened 2018-08-02
i18n: added Chinese-Simple translation

Steinkuo:
added Chinese-Simple translation


Steinkuo:
I found that zh_CN and nl_NL translation is not working properly. I check it.
The file name is changed to flask_security and can be translated normally.
 I hope to help you.

jirikuncar (approved):
Thanks!

PR #809, opened 2018-10-26
Use get_identity_attributes instead of email in role commands

yoursdearboy:


jwag956:
I think I see what you want to do - but I don't think this is quite the right approach - since relying on ordering in get_identity_attributes seems fragile.
I would propose supporting sending in 'user' as a dict {<identity_attribute>: value} and issuing find_user with that...

PR #858, opened 2020-02-12
Use new Flask-Login request_loader to set request callback

### LINKED ISSUE #856: Flask-Login dependency
Flask-Login has released v0.5.0 on 2020-02-09.

This release has broken some of the decorators (all I have encountered, but not necessarily exclusive to decorators).
```
File "/usr/local/lib/python3.6/site-packages/flask_security/decorators.py", line 61, in _check_token
    user = _security.login_manager.request_callback(request)
AttributeError: 'LoginManager' object has no attribute 'request_callback'
```
A more specific version check might be needed:
https://github.com/mattupstate/flask-security/blob/develop/setup.py#L50

For those having the same issue add a fixed dependency to `Flask-Login` <= 0.5.0
e.g.:
```
Flask-Login==0.4.1
```
(0.4.1 was the greatest version < 0.5.0 at the time of writing)

I will check back with more details (and further testing) once I find time and perhaps submit a PR.

alanhamlett:
Request loader callback was renamed in Flask-Login 0.5.0 to conform with other user/header loader names. Fixes #856.

### DIFF
--- a/flask_security/decorators.py
+++ b/flask_security/decorators.py
@@ -63,7 +63,7 @@ def _get_unauthorized_view():
 
 
 def _check_token():
-    user = _security.login_manager.request_callback(request)
+    user = _security.login_manager.request_loader(request)
 
     if user and user.is_authenticated:
         app = current_app._get_current_object()


flask_security/decorators.py
```diff
@@ -63,7 +63,7 @@ def _get_unauthorized_view():
 
 
 def _check_token():
-    user = _security.login_manager.request_callback(request)
+    user = _security.login_manager.request_loader(request)
```
smurf-U, line 66: `request_loader` should return `User` object but it returns `LocalProxy`. Because of that, while executing the next line it gives the error that below error.

> AttributeError: 'Request' object has no attribute 'is_authenticated'

If you replace `request_loader `  with `_request_callback` it gives proper output. But it's a wrong way because it is a static method.

jwag956: Actually - they made it python 'local' - and is called _request_callback now. I have requested that they restore that to public - which they did by providing some accessor properties. That is merged but not released.

smurf-U: > Actually - they made it python 'local' - and is called _request_callback now. I have requested that they restore that to public - which they did by providing some accessor properties. That is merged but not released.

Any ETA for release.

